PR #8, opened 2019-04-18
Use Set instead of when collecting project dependencies

geoand:
This is done to avoid having duplicates due to the multitude of Gradle
configurations

### DIFF
--- a/analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentService.java
+++ b/analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentService.java
@@ -1,6 +1,6 @@
 package org.jboss.gm.analyzer.alignment;
 
-import java.util.List;
+import java.util.Collection;
 
 import org.commonjava.maven.atlas.ident.ref.ProjectVersionRef;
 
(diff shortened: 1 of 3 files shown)


analyzer/src/main/java/org/jboss/gm/analyzer/alignment/DummyAlignmentService.java
```diff
@@ -24,7 +24,7 @@ private String getProjectAlignment(Request request) {
             return result;
         }
 
-        final ProjectVersionRef first = request.getDependencies().get(0);
+        final ProjectVersionRef first = request.getDependencies().iterator().next();
```
geoand: We shouldn't worry about this for now, since in my next PR this will be going away.

metacosm: OK, what will you use instead? I worry about relying on mixing project and dependencies in the same bag…

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentService.java
```diff
@@ -9,15 +9,15 @@
     Response align(Request request);
 
     class Request {
-        private final List<? extends ProjectVersionRef> dependencies;
+        private final Collection<? extends ProjectVersionRef> dependencies;
```
metacosm: Might as well specify Set semantics?

geoand: Not sure... We might have to change this again in the future so I would rather keep it generic for now, what do you think?

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/DummyAlignmentService.java
geoand: `DummyAlignmentService` is going to be deleted

PR #11, opened 2019-04-18
Implement dependencies override plugin

metacosm:


geoand:
Is this still a draft? If so, what functionality is missing from it?

geoand:
I guess one thing that needs to be added is support for the old `maven` plugin (in addition to `maven-publish`).

manipulation/src/main/java/org/jboss/gm/manipulation/ManipulationPlugin.java
```diff
@@ -16,11 +17,12 @@
 
     @Override
     public void apply(Project project) {
-        final String projectName = project.getName();
+        // apply the maven publishing plugin
+        project.getPluginManager().apply(MavenPublishPlugin.class);
```
rnc: What happens if the project already applies the mavenpublishplugin - do any gradle projects, expecting to be deployed, *not* put it in?

manipulation/src/main/java/org/jboss/gm/manipulation/actions/ManifestUpdateAction.java
```diff
@@ -0,0 +1,40 @@
...
+import org.gradle.api.tasks.bundling.Jar;
+import org.jboss.gm.common.alignment.AlignmentModel;
+
+/**
+ * Overrides specified OSGI manifest entries.
+ * <p>
+ * Currently only overrides manifest of type OsgiManifest and only entries "Specification-Version" and
+ * "Implementation-Version".
+ */
+public class ManifestUpdateAction implements Action<Project> {
```
rnc: Why do we need this?

manipulation/src/main/java/org/jboss/gm/manipulation/actions/PomTransformer.java
```diff
@@ -0,0 +1,91 @@
...
+import org.gradle.api.XmlProvider;
+import org.jboss.gm.common.alignment.AlignmentModel;
+import org.w3c.dom.Element;
+import org.w3c.dom.Node;
+import org.w3c.dom.NodeList;
+
+/**
+ * Overrides data in pom generated by the old 'maven' plugin.
+ */
+public class PomTransformer implements Action<XmlProvider> {
```
rnc: If a Gradle project generates pom files for deployment, does it retrieve the information from the original unmodified build.gradle or from the in-memory altered model?

manipulation/src/main/java/org/jboss/gm/manipulation/actions/PublishingRepositoryAction.java
```diff
@@ -0,0 +1,67 @@
...
+ */
+public class PublishingRepositoryAction implements Action<Project> {
+
+    public static final String URL_SYSTEM_PROPERTY = "AProxDeployUrl";
+
+    public static final String ACCESS_TOKEN_SYSTEM_PROPERTY = "accessToken";
+
+    @Override
+    public void execute(Project project) {
+        String pncDeployUrl = System.getProperty(URL_SYSTEM_PROPERTY);
```
rnc: Are the system properties populated with environment variables? - I believe AProxDeployUrl is an env variable in the pod.

rnc:
Looks great, added a couple of questions :-)

manipulation/src/main/java/org/jboss/gm/manipulation/ManipulationPlugin.java
geoand: There are projects that use the older `maven` plugin so we will need to take care of that case as well.

Also, applying pluggins is idempotend according to: https://docs.gradle.org/current/userguide/plugins.html#sec:using_plugins

manipulation/src/main/java/org/jboss/gm/manipulation/actions/ManifestUpdateAction.java
geoand: Does PME update the generated Manifests? If so, then I would assume we would want it here as well.

The only reason I can think of to have the manifests populated with this info is for `getClass().getPackage().getImplementationVersion` to return the updated version

manipulation/src/main/java/org/jboss/gm/manipulation/actions/PublishingRepositoryAction.java
geoand: I am starting to think that the plugin needs to check whether `maven` or `maven-publish` plugin is configured and act accordingly. If none is configured (which I am assuming is very unlikely), we can just choose one.

manipulation/src/main/java/org/jboss/gm/manipulation/actions/UploadTaskTransformerAction.java
```diff
@@ -0,0 +1,29 @@
...
+
+    public UploadTaskTransformerAction(AlignmentModel.Module alignmentConfiguration) {
+        this.alignmentConfiguration = alignmentConfiguration;
+    }
+
+    @Override
+    public void execute(Project project) {
+        project.getTasks().withType(Upload.class).all(upload -> upload.getRepositories()
+                .withType(MavenResolver.class).all(resolver -> {
+                    resolver.getPom().withXml(new PomTransformer(alignmentConfiguration));
```
geoand: I am wondering if this is the best approach (same for `PublicationPomTransformerAction`)... We could also do ` resolver.getPom().whenConfigured(...)` and then work with the `MavenPom` type which would be much easier than playing with XML. WDYT?

geoand: Actually it seems like the snippet I pasted above would only work for `UploadTaskTransformerAction` and not for `PublicationPomTransformerAction`

manipulation/src/functTest/java/org/jboss/gm/manipulation/TestUtils.java
```diff
@@ -16,4 +30,57 @@ public static void copyDirectory(String classpathResource, File target) throws U
         FileUtils.copyDirectory(Paths
                 .get(TestUtils.class.getClassLoader().getResource(classpathResource).toURI()).toFile(), target);
     }
+
+    static Tuple getAlignedTuple(AlignmentModel.Module alignment, String artifactId, String expected) {
```
geoand: Good stuff!

manipulation/src/functTest/java/org/jboss/gm/manipulation/MultiModuleFunctionalTest.java
```diff
@@ -40,53 +40,38 @@ public void ensureProperPomGenerated() throws IOException, URISyntaxException, X
...
-        final Model model = reader.read(new FileReader(generatedPomPath.toFile()));
-        assertThat(model.getGroupId()).isEqualTo("org.acme");
-        assertThat(model.getArtifactId()).isEqualTo("subproject1");
-        assertThat(model.getVersion()).isEqualTo("1.1.2");
-        assertThat(model.getDependencies())
+    private void assertSubproject1Pom(File multiModuleRoot, AlignmentModel alignment) throws IOException,
+            XmlPullParserException {
+        final Pair<Model, AlignmentModel.Module> modelAndModule = TestUtils.getModelAndCheckGAV(multiModuleRoot, alignment,
+                "subproject1/build/publications/main/pom-default.xml");
+        final AlignmentModel.Module module = modelAndModule.getRight();
```
geoand: Maybe we should do something like the following in order to test the generated version as well:

```
        final AlignmentModel.Module module = modelAndModule.getRight();
        assertThat(modelAndModule.getLeft()).satisfies(model -> {
            assertThat(model.getVersion()).isEqualTo(module.getNewVersion());
            assertThat(model.getDependencies())
                    .extracting("artifactId", "version")
                    .containsOnly(
                            TestUtils.getAlignedTuple(module, "hibernate-core"),
                            TestUtils.getAlignedTuple(module, "commons-lang3", "3.8.1"),
                            TestUtils.getAlignedTuple(module, "undertow-core"));
        });
```

manipulation/src/main/java/org/jboss/gm/manipulation/actions/ManifestUpdateAction.java
rnc: No PME doesn't because change the manifest explicitly, because generation uses the updated model.

geoand: But ultimately the changes to the model are reflected in the manifest, correct? I think that is not the case for Gradle (although I'm not certain)

manipulation/src/main/java/org/jboss/gm/manipulation/ManipulationPlugin.java
rnc: Ok but what happens if there is both maven and maven publish plugins?

manipulation/src/main/java/org/jboss/gm/manipulation/actions/ManifestUpdateAction.java
rnc: Yes, changes to the model are reflected in the pom files, the manifests etc.

I am not sure with Gradle hence asking about this and the poms :-)

manipulation/src/main/java/org/jboss/gm/manipulation/actions/PublishingRepositoryAction.java
rnc: Could we just remove any existing instance / configuration and replace with our known configuration?

manipulation/src/main/java/org/jboss/gm/manipulation/actions/ManifestUpdateAction.java
TomasHofman: Yes, for maven projects built in PNC the versions in the manifest are the updated versions (because projects are built with updated pom.xml). So this mimics maven project behavior.

I'm not sure if it's critical to have this, but it seems a good idea and it's not difficult. I thought that some customer automation could theoretically rely on this...

manipulation/src/main/java/org/jboss/gm/manipulation/actions/PomTransformer.java
TomasHofman: That depends on how the pom is generated. I think the most common is using "maven" plugin or "maven-publish" plugin. IIRC both of them use unresolved (i.e. the original) dependency versions, so that's why this transformer.

manipulation/src/main/java/org/jboss/gm/manipulation/actions/PomTransformer.java
```diff
@@ -0,0 +1,91 @@
...
+import org.commonjava.maven.atlas.ident.ref.ProjectVersionRef;
+import org.gradle.api.Action;
+import org.gradle.api.XmlProvider;
+import org.jboss.gm.common.alignment.AlignmentModel;
+import org.w3c.dom.Element;
+import org.w3c.dom.Node;
+import org.w3c.dom.NodeList;
+
+/**
+ * Overrides data in pom generated by the old 'maven' plugin.
```
TomasHofman: This comment is in fact wrong, because this transformer is used also for the maven-publish plugin.

manipulation/src/main/java/org/jboss/gm/manipulation/actions/PublishingRepositoryAction.java
TomasHofman: > Are the system properties populated with environment variables? - I believe AProxDeployUrl is an env variable in the pod.

I don't think so, I set it via -D switch in the gradle command. But it would be handy improvement - first looks for system prop and then fallback to `System.getenv()`?

TomasHofman: > I am starting to think that the plugin needs to check whether `maven` or `maven-publish` plugin is configured and act accordingly. If none is configured (which I am assuming is very unlikely), we can just choose one.

+1

manipulation/src/main/java/org/jboss/gm/manipulation/ManipulationPlugin.java
metacosm: I haven't checked that use case yet.

manipulation/src/main/java/org/jboss/gm/manipulation/actions/ManifestUpdateAction.java
metacosm: 👍

manipulation/src/main/java/org/jboss/gm/manipulation/actions/UploadTaskTransformerAction.java
metacosm: We can certainly re-visit things as we see fit.

manipulation/src/main/java/org/jboss/gm/manipulation/ManipulationPlugin.java
geoand: Me neither

manipulation/src/main/java/org/jboss/gm/manipulation/actions/PomTransformer.java
geoand: When I was trying things I also saw that the pom files did not reflect dependency updated the changes OOTB

manipulation/src/main/java/org/jboss/gm/manipulation/actions/PublishingRepositoryAction.java
geoand: > Could we just remove any existing instance / configuration and replace with our known configuration?

I am not sure, we'll need to check this.

geoand: > Are the system properties populated with environment variables? - I believe AProxDeployUrl is an env variable in the pod.

We should probably do something similar to what is being done in the other plugin where Apache Commons Config is used in order to read config from both system properties and environment variables.

manipulation/src/main/java/org/jboss/gm/manipulation/ManipulationPlugin.java
rnc: Ok, no problem, we need to log these thoughts in jira I think

geoand:
retest this please

PR #20, opened 2019-04-24
Add publishing for common

rnc:


common/build.gradle.kts
```diff
@@ -1,8 +1,12 @@
-plugins {
-    java
-}
+group = "org.jboss.gm.common"
 
-version = "0.1-SNAPSHOT"
+publishing {
+    publications {
+        register("mavenJava", MavenPublication::class) {
```
metacosm: Why is this needed?

rnc: Trying to get the common module published so that the plugins can use it

PR #22, opened 2019-04-24
Improvements

metacosm:


common/src/main/java/org/jboss/gm/common/alignment/AlignmentUtils.java
```diff
@@ -5,12 +5,14 @@
...
 import java.nio.file.Path;
+import java.util.HashMap;
+import java.util.Map;
 
 import org.apache.commons.io.FileUtils;
 
 public final class AlignmentUtils {
-
     private static final String ALIGNMENT_FILE_NAME = "alignment.json";
+    private static final Map<String, ManipulationModel> cachedModels = new HashMap<>(7);
```
geoand: Any particular reason you chose `7` :) ?

metacosm: Small non-pair, prime number which used to be a good idea to use for an initial Map size according to Effective Java. Not sure that's still the case.

geoand: Great to know!

metacosm:
test this please

geoand:
@metacosm do you prefer to squash this or not? I am leaning on no because there seem to be multiple themes here

metacosm:
Agreed.

PR #26, opened 2019-04-25
feat: configure maven-publish plugin publications

TomasHofman:
* work towards using maven or maven-publish plugin for publishing to PNC, depending on which plugin is used by the project
* removes license headers

### DIFF
--- a/common/src/main/java/org/jboss/gm/common/Configuration.java
+++ b/common/src/main/java/org/jboss/gm/common/Configuration.java
@@ -48,6 +48,9 @@ public interface Configuration extends Accessible {
     @Key("AProxDeployUrl")
     String deployUrl();
 
+    @Key("deployPlugin")
+    String deployPlugin();
+
     @Key("accessToken")
     String accessToken();
 
--- a/manipulation/src/main/java/org/jboss/gm/manipulation/ManipulationPlugin.java
+++ b/manipulation/src/main/java/org/jboss/gm/manipulation/ManipulationPlugin.java
@@ -1,18 +1,33 @@
 package org.jboss.gm.manipulation;
 
+import static org.apache.commons.lang3.StringUtils.isEmpty;
 import static org.jboss.gm.common.alignment.ManipulationUtils.getCurrentManipulationModel;
 
+import org.aeonbits.owner.ConfigCache;
+import org.gradle.api.InvalidUserDataException;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
+import org.gradle.api.plugins.MavenPlugin;
+import org.gradle.api.publish.maven.plugins.MavenPublishPlugin;
+import org.jboss.gm.common.Configuration;
 import org.jboss.gm.common.alignment.ManipulationModel;
 import org.jboss.gm.manipulation.actions.ManifestUpdateAction;
+import org.jboss.gm.manipulation.actions.MavenPublicationRepositoryAction;
 import org.jboss.gm.manipulation.actions.OverrideDependenciesAction;
 import org.jboss.gm.manipulation.actions.ProjectChangeVersionAction;
-import org.jboss.gm.manipulation.actions.PublicationPomTransformerAction;
+import org.jboss.gm.manipulation.actions.PublishingArtifactsAction;
+import org.jboss.gm.manipulation.actions.PublishingPomTransformerAction;
 import org.jboss.gm.manipulation.actions.PublishingRepositoryAction;
 import org.jboss.gm.manipulation.actions.UploadTaskTransformerAction;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class ManipulationPlugin implements Plugin<Project> {
+
+    private static final String LEGACY_MAVEN_PLUGIN = "maven";
+    private static final String MAVEN_PUBLISH_PLUGIN = "maven-publish";
+
+    private static final Logger log = LoggerFactory.getLogger(ManipulationPlugin.class);
 
     @Override
     public void apply(Project project) {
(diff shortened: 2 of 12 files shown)


manipulation/src/main/java/org/jboss/gm/manipulation/actions/MavenPublicationRepositoryAction.java
```diff
@@ -0,0 +1,54 @@
...
+            return;
+        }
+
+        String url = (String) project.getProperties().get(URL_PROPERTY);
+        if (isEmpty(url)) {
+            log.warn("{} project property is empty, publication repository will not be configured.", URL_PROPERTY);
+            return;
+        }
+
+        String accessToken = (String) project.getProperties().get(ACCESS_TOKEN_PROPERTY);
```
rnc: Consider using the ConfigurationFactory

rnc:
Looks good!

Can we xref the NCL-4614 ticket in the commit message please?

manipulation/src/main/java/org/jboss/gm/manipulation/ManipulationPlugin.java
```diff
@@ -23,9 +32,45 @@ public void apply(Project project) {
...
-        project.afterEvaluate(new UploadTaskTransformerAction(correspondingModule));
-        project.afterEvaluate(new PublicationPomTransformerAction(correspondingModule));
+        //        project.afterEvaluate(new UploadTaskTransformerAction(correspondingModule));
+        //        project.afterEvaluate(new PublishingPomTransformerAction(correspondingModule));
+
+        // TODO: add functional tests
+        project.afterEvaluate(new Action<Project>() {
+            @Override
+            public void execute(Project project) {
+                Object preferenceProperty = project.getProperties().get("gm.publishing-plugin");
```
rnc: Where does this come from please? What happens if its not set?

manipulation/src/main/java/org/jboss/gm/manipulation/ManipulationPlugin.java
```diff
@@ -23,9 +32,45 @@ public void apply(Project project) {
...
+                    project.getPluginManager().apply(MavenPlugin.class);
+                } else if ("maven-publish".equals(preferenceProperty)
+                        && !project.getPluginManager().hasPlugin("maven-publish")) {
+                    log.info("Applyng `maven-publish` plugin.");
+                    project.getPluginManager().apply(MavenPublishPlugin.class);
+                    project.afterEvaluate(new PublishingArtifactsAction());
+                }
+
+                // determine which publishing plugin to use
+                if (project.getPluginManager().hasPlugin("maven")) {
```
rnc: What happens if the project is already configured with e.g. `maven` - do we need to remove the existing configurations?

manipulation/src/main/java/org/jboss/gm/manipulation/actions/AlignedDependencyResolver.java
```diff
@@ -43,6 +49,11 @@ public AlignedDependencyResolver(ManipulationModel module) {
     public void execute(DependencyResolveDetails resolveDetails) {
         if (!BYPASS) {
             final ModuleVersionSelector requested = resolveDetails.getRequested();
+            if (isEmpty(requested.getVersion())) {
```
rnc: Is this a similar problem to the getAllProjectDependencies which also needed a similar check ?

manipulation/src/main/java/org/jboss/gm/manipulation/actions/MavenPublicationRepositoryAction.java
rnc: Ignore this - would be solved by NCL-4632

manipulation/src/main/java/org/jboss/gm/manipulation/ManipulationPlugin.java
TomasHofman: So, I was trying to use a "project property", because I thought I had problem that "system properties" were not correctly set when the gradle daemon was already running, but I can't reproduce it anymore. So I'm going to switch back to system property or will use our ConfigurationFactory.

As for the meaning of the property, the intention was that we would reuse the publishing plugin used by the project (which should be easier for us, we would basically only need to add our publication repo, and the remaining configuration - artifacts to publish, poms, etc - should be already configured by the project). This property is meant to override this behavior.

Btw project properties can be set:
* in project's gradle.properties file, without prefix, ie `gm.publishing-plugin=xxx`,
* via -P switch in the `gradle` command.

Contrary to that "system properties" (obtained via `System.getProperty()`) can be set:
* in gradle.properties files, with "systemProp." prefix, ie `systemProp.gm.publishing-plugin=xxx`
* via -D switch.

manipulation/src/main/java/org/jboss/gm/manipulation/ManipulationPlugin.java
TomasHofman: As I described above, I want to intentionally use the already configured plugin, which simplifies configuration for us. Possibly we may need to remove only the existing publication repository config, but not in Spring case, because it doesn't have any repo configured. In Hibernate I didn't need to do that either.

manipulation/src/main/java/org/jboss/gm/manipulation/actions/AlignedDependencyResolver.java
TomasHofman: I think you run into this yesterday - this is to avoid "managed dependencies". The logic bellow fails when the version is empty.

TomasHofman: Will add comment.

manipulation/src/main/java/org/jboss/gm/manipulation/ManipulationPlugin.java
rnc: Ok. I've added a new Configuration class instead of a ConfigFactory.

manipulation/src/main/java/org/jboss/gm/manipulation/ManipulationPlugin.java
rnc: Ok thanks

manipulation/src/main/java/org/jboss/gm/manipulation/actions/AlignedDependencyResolver.java
rnc: Ok thanks

common/src/main/java/org/jboss/gm/common/Configuration.java
```diff
@@ -48,6 +48,9 @@ default Properties getProperties() {
     @Key("AProxDeployUrl")
     String deployUrl();
 
+    @Key("deployPlugin")
+    String deployPlugin();
```
rnc: From the look of the usage deployPlugin only has two valid values? So why not make it a enum for improved static typing?

manipulation/src/main/java/org/jboss/gm/manipulation/actions/PublishingRepositoryAction.java
```diff
@@ -45,26 +41,35 @@
...
+        project.getPlugins().withType(MavenPublishPlugin.class, plugin -> {
+            project.getExtensions().configure(PublishingExtension.class, publishingExtension -> {
+                publishingExtension.getRepositories().maven(repository -> {
+                    repository.setName("Manipulator Publishing Repository");
+                    repository.setUrl(config.deployUrl());
+                    if (!isEmpty(config.deployUrl())) {
                         repository.credentials(HttpHeaderCredentials.class, cred -> {
                             cred.setName("Authorization");
-                            cred.setValue("Bearer " + config.accessToken());
+                            cred.setValue("Bearer " + System.getProperty(config.accessToken()));
```
rnc: Can you explain this please? Shouldn't config.accessToken return the value?

TomasHofman: Oh sure, mistake during copy-paste :).

common/src/main/java/org/jboss/gm/common/Configuration.java
rnc: ( Discussed offline ; we'll leave this for now ).

PR #28, opened 2019-04-29
NCL-4632 Replace commons-configuration with OWNER 

rnc:
Draft for early feedback. Note - need the manipulation functional tests fixing - discussing with @metacosm what the test is meant to be doing).


metacosm:
#29 brings the modification of the tests we talked about that removes the need to bypass the dependencies override to make the test pass.

rnc:
@metacosm Ah damn, I've just done the same as well!

metacosm:
@rnc  you need to do the same for the multi-module test as well and remove the bypass property set: https://github.com/project-ncl/gradle-manipulator/blob/master/manipulation/build.gradle.kts#L40

rnc:
@metacosm I've cherry-picked your commit onto this branch :-) Thanks!

rnc:
I think this is ready for review now. I'd rather it was merged than squashed and merged to preserve the commits.

PR #30, opened 2019-05-01
NCL-4587 Add injection for automatic usage of the ManipulationPlugin

rnc:
This is in draft as its waiting on #26 by @TomasHofman  to be completed.

### DIFF
--- a/analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
+++ b/analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
@@ -4,6 +4,12 @@ import static org.apache.commons.lang3.StringUtils.isEmpty;
 import static org.jboss.gm.common.alignment.ManipulationUtils.getCurrentManipulationModel;
 import static org.jboss.gm.common.alignment.ManipulationUtils.writeUpdatedManipulationModel;
 
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.StandardCopyOption;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.LinkedHashSet;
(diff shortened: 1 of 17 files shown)


analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
```diff
@@ -58,7 +67,45 @@ public void perform() {
 
         projectsToAlign.remove(projectName);
         if (projectsToAlign.isEmpty()) {
+
+            System.out.println("#### settings is " + project.getGradle().getRootProject() + " and "
```
metacosm: Should we keep this output?

analyzer/src/main/resources/gme.gradle
```diff
@@ -0,0 +1,22 @@
...
+    }
+
+    dependencies {
+        classpath "org.jboss.gm.manipulation:manipulation:${project.version}"
+    }
+}
+
+// If we are not running the analyse plugin then apply the manipulation plugin.
+// This avoids issues with rerunning in local environments.
+if (!project.hasProperty("gmeAnalyse")) {
```
metacosm: Where does that come into play? I don't see any reference to this property anywhere (though I might just have missed it)…

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
rnc: Damn. Left over debugging! I'll remove.

analyzer/src/main/resources/gme.gradle
rnc: Its linked to my init.gradle which sets the value to true. 
I'll create a separate ticket for this as I think we should store the init.gradle in the repo somewhere so we can discuss and have things like
```
allprojects {
    ext.gmeAnalyse = true
    apply plugin: org.jboss.gm.analyzer.alignment.AlignmentPlugin
```

metacosm: I agree about storing the `init.gradle` in the repo: I had actually started working on it at some point.

PR #33, opened 2019-05-03
NCL-4658 Add init gradle file

rnc:
Directory structure can be discussed and whether to add my `init.iml` that I use for configuring this sub-module (it basically sets up groovy and adds gradle to the classpath).

### DIFF
--- a/init/src/analyzer.init.gradle
+++ b/init/src/analyzer.init.gradle
@@ -1,0 +1,33 @@
+
+
+initscript {
+    repositories {
+        mavenLocal()
+        mavenCentral()
+    }
+    dependencies {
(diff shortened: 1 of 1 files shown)


rnc:
@metacosm @geoand Couple of points with this - I would rather we had a better way of updating the version? Not sure if we can make it part of a release process? Secondly, the logging needs work so not sure if this should be done separately or added as a conditional?

geoand:
@rnc Have any of us looked into how releases of Gradle plugins is typically done or any examples from other plugins?

rnc:
@geoand Yeah I think we should either publish to gradle portal and/or maven central. That is probably separate to this issue though ( and I think there is an existing release jira for it).

geoand:
@rnc Makes sense. 

This looks good to me!

PR #37, opened 2019-05-07
Use resolved dependencies when no version is present

geoand:


geoand:
test this please

geoand:
retest this please

geoand:
I'm going to close this since this approach won't work at the end of the process, because the generated poms from publishing process will still import the BOM which will override whatever version we set.

PR #41, opened 2019-05-08
NCL-4657 Make plugins work with Spring Dependency Management Plugin

geoand:
This support only works when the the plugin is configured to NOT
generate a dependencyManagement section in the generated pom (like is
currently the case for Spring Framework)

### DIFF
--- a/manipulation/src/main/java/org/jboss/gm/manipulation/ManipulationPlugin.java
+++ b/manipulation/src/main/java/org/jboss/gm/manipulation/ManipulationPlugin.java
@@ -3,7 +3,10 @@ package org.jboss.gm.manipulation;
 import static org.apache.commons.lang3.StringUtils.isEmpty;
 import static org.jboss.gm.common.alignment.ManipulationUtils.getCurrentManipulationModel;
 
+import java.lang.reflect.Method;
+
 import org.aeonbits.owner.ConfigCache;
+import org.commonjava.maven.ext.common.ManipulationUncheckedException;
 import org.gradle.api.InvalidUserDataException;
 import org.gradle.api.Plugin;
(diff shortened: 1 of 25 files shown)


manipulation/src/main/java/org/jboss/gm/manipulation/ManipulationPlugin.java
```diff
@@ -40,18 +44,51 @@ public void apply(Project project) {
...
 
-        configurePublishingTask(project, correspondingModule);
+        configurePublishingTask(project, correspondingModule, resolvedDependenciesRepository);
+    }
+
+    // Ensure that if the Spring Dependency Management plugin is applied,
+    // that it's configured to not generate a "dependencyManagement" section in the generated bom
+    // This is needed because if we don't do it, the "dependencyManagement" section (which is a bom inclusion) will override our dependencies
+    // On the implementation side of things, we need to use reflection because we can get a nasty classloader errors
+    // when trying to cast the object to the known type
```
iocanel: Can you provide some more info on the class loading errors you get?

geoand: Sure thing. Basically I can't cast to the target class because it gets loaded by a different class loader than this class.
If you have a better suggestion than reflection, I would LOVE to follow it

rnc: Is it similar to the SLF4J / MDC problem ?

geoand: It seemed very similar yes

PR #48, opened 2019-05-14
Ensure that we skip unresolvable project modules

geoand:


rnc:
( Suggested alternative implementation in offline discussion )

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
```diff
@@ -118,17 +123,27 @@ private void writeMarkerFile(File rootDir) throws IOException {
 
             if (configuration.isCanBeResolved()) {
 
+                final Set<ProjectRef> allProjectDependencies = configuration.getAllDependencies()
```
rnc: Should this not be getDependencies() rather than getAllDependencies?

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
```diff
@@ -118,17 +123,27 @@ private void writeMarkerFile(File rootDir) throws IOException {
...
+                        .stream()
+                        .filter(d -> ProjectDependency.class.isAssignableFrom(d.getClass()))
+                        .map(d -> new SimpleProjectRef(d.getGroup(), d.getName()))
+                        .collect(Collectors.toSet());
```
rnc: Why not just store the project name - the String ?

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
geoand: I tried `getDependencies()` and was getting 0 dependencies back for the new test case.

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
```diff
@@ -150,6 +165,18 @@ private void writeMarkerFile(File rootDir) throws IOException {
...
 
+    private Set<SimpleProjectRef> getUnresolvedDependenciesExcludingProjectDependencies(LenientConfiguration lenient,
+            Set<ProjectRef> allProjectModules) {
+        return lenient.getUnresolvedModuleDependencies()
+                .stream()
+                .map(d -> {
+                    final ModuleVersionSelector selector = d.getSelector();
+                    return new SimpleProjectRef(selector.getGroup(), selector.getName());
+                })
+                .filter(p -> !allProjectModules.contains(p))
```
rnc: If we are passing in Set<String> projectDependencies then I think this could be made simpler?

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
geoand: Instead of using both the group and name? 
I guess I just want to be caution and make it work no matter what. 
I could just use the name if you think it's safe.

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
geoand: I thought so initially as well, but `Dependency` does not related to `UnresolvedDependency`, so you can't really use them for comparison (unless I missed something)

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
rnc: Ok (discussed offline). Perhaps add a comment for future reference.

geoand: :+1:, comment added

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
rnc: Ah yes good point. If we're being really safe, we should store the version as well. How about making it a set of ProjectDependencies - avoids the new SimpleProjectRef?

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
rnc: Can we use the moduleversionselector to compare with the projectdependency group:name:version?

rnc: UnresolvedDependency.getSelector?

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
geoand: We don't reallt need to check on the version do we?

Good point about `SimpleProjectRef`. But if we go down that route, then using the `Set` won't do us any good, because we won't be able to compare (via `equals`) the `Dependency` with the `UnresolvedDependency` - or perhaps you have seen something I haven't?

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
geoand: Searched for something, came up empty :(

PR #52, opened 2019-05-15
NCL-4663 Configure publishing of the plugins to the portal

geoand:
The configuration follows the standard gradle suggestions at:
https://guides.gradle.org/publishing-plugins-to-gradle-plugin-portal/


rnc:
Can we add:
- displayName/description to the plugin block.
- documentation on the release process

Questions:

- Will this also release the javadoc/sources jars?
- While I suspect this isn't required, do we also need to consider releasing to maven central ?

geoand:
@rnc I will add the sections you mention and some data about the release process (although it will be untested until we actually do a release).

As for the questions, I would assume the first is correct (but again won't know for sure until we do a release). For Maven Central I don't think we need this ATM. Do you have anything specific reason in mind why it would be helpful?

PR #55, opened 2019-05-16
Handle "caching" of data between tasks without relying on static

geoand:


analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AdditionalPropertiesManipulationModelCache.java
```diff
@@ -0,0 +1,35 @@
...
+        try {
+            return (ManipulationModel) AdditionalPropertiesUtil.getProperty(project, NAME_PREFIX + name);
+        } catch (ExtraPropertiesExtension.UnknownPropertyException e) {
+            return null;
+        }
+    }
+
+    @Override
+    public void put(String name, ManipulationModel manipulationModel) {
+        AdditionalPropertiesUtil.setPropertyIfMissing(project, NAME_PREFIX + name, manipulationModel);
```
TomasHofman: Should we allow overriding? I can imagine being surprised my value not being stored if the property already existed.

geoand: Indeed, but this is very internal and pretty much only used here. I don't mind adding it however

PR #59, opened 2019-05-17
NCL-4711 Test case verifies that internal project deps are not includ…

TomasHofman:
…ed in manipulation.json


geoand:
Great addition to the test suite!

PR #60, opened 2019-05-17
Avoid resolution errors

rnc:


geoand:
Would it be possible to have a test that fails without this change or is it too difficult to create?

PR #63, opened 2019-05-23
Artifact publishing doc & stylesheet fix

TomasHofman:


guide/dep-manip.md
```diff
@@ -163,3 +163,7 @@ This will prevent any alignment within the org.foo:moduleB.
 
 This will prevent any alignment within org.foo and all sub-modules within that.
 
+#### Disabling Dependency Manipulation
+
+TODO: Introduce new parameter? Setting -DdependencySource=NONE is not sufficient, because that also disables alignment 
+of the project version.
```
rnc: You're right it does disable alignment - it still produces a redhat-1 but without source information it can't know whether it should increment it past 1. Potentially in future we could also look at any existing manipulation.json to help determine the redhat version

PR #64, opened 2019-05-24
Add note about the init scripts inside the project

geoand:


geoand:
Of course once we release the plugin a link should be added to show where the init script can be found

rnc:
I would agree with adding a link once we do the first release. Possibly also in `contributing` we could mention how to get the work-in-progress init script?

geoand:
@rnc that's a great idea!

PR #69, opened 2019-05-28
Fix publishing setup

metacosm:
Based on https://guides.gradle.org/publishing-plugins-to-gradle-plugin-portal/ but https://plugins.gradle.org/docs/publish-plugin seems to indicate that description can be in `pluginBundle` block… so 🤷‍♂


metacosm:
Or we could just leave everything as-is 😄

geoand:
Your call :)

PR #70, opened 2019-05-28
NCL-4726 Export artifact repository settings

TomasHofman:
https://projects.engineering.redhat.com/browse/NCL-4726

This should probably wait until after the first release.

Functionality:

If `-DrepoRemovalBackup=path/to/settings.xml` property is configured, `settings.xml` file will be generated and will contain remote repositories used by the project. This file will be picked by PNC in order to automatically configure repository proxying. (I hope I got this right, that is what PME does.)

The property name `repoRemovalBackup` is the same as in PME (respectively PME is accepting two variants, camel case or dashed).


geoand:
Cool! I can't really comment since I don't have the PME experience, but this looks neat

TomasHofman:
Cc @rnc when you have time, not urgent...

rnc:
+1 thanks for implementing that idea.

cc @vibe13 @thescouser89  -  this should help and also potentially reduce any work @pkocandr has to do.

rnc:
@TomasHofman Can we also get some updates for the documentation please? (on the gh-pages branch)

PR #72, opened 2019-05-28
NCL-4663 Another attempt to get shadow/maven-publication working

rnc:
@geoand @metacosm  @TomasHofman  This isn't working correctly at the moment :-( I have not been able to get it to call `gradle release -Drelease=true` and either implicitly call `publish` or explicitly call it via `gradle release publish -Drelease=true` **and** maintain the correct order. Unfortunately it seems to try to deploy to sonatype before tagging and versioning the release.


build.gradle.kts
```diff
@@ -6,13 +6,42 @@ plugins {
...
     id("net.researchgate.release") version "2.6.0"
 }
 
+apply(plugin = "net.researchgate.release")
+release {
+    failOnUnversionedFiles = false
+    failOnCommitNeeded = false
+    failOnPublishNeeded = false
+    failOnSnapshotDependencies = false
+    failOnUnversionedFiles = false
```
rnc: This should all be removed once its working.

build.gradle.kts
```diff
@@ -6,13 +6,42 @@ plugins {
...
+    )
+}
+tasks.named<Task>("afterReleaseBuild") {
+    println ("### Now invoking publish...")
+
+    dependsOn (
+            "analyzer:publishShadowPublicationToSonatype-nexus-stagingRepository",
+            "manipulator:publishShadowPublicationToSonatype-nexus-stagingRepository"
+            //    dependsOn( project.tasks.findByPath("shadow") )
+            //    dependsOn( "publish" )
```
rnc: I think the problem is here.

geoand:
@rnc were would you like me to start looking?

rnc:
@geoand Sent you message on hangouts.

geoand:
I am fairly certain that only extra bit of configuration need for the release plugin is:

```
tasks.afterReleaseBuild {
    dependsOn(":analyzer:publish", ":manipulation:publish")
}
```

build.gradle.kts
```diff
@@ -159,6 +162,12 @@ subprojects {
                     }
                 }
             }
+            repositories {
+                maven {
+                    name = "sonatype-nexus-staging"
```
geoand: Are we keeping staging after all?

gradle.properties
```diff
@@ -1 +1 @@
-version=0.1-SNAPSHOT
\ No newline at end of file
+version=0.1-SNAPSHOT
```
geoand: Should this be reverted?

build.gradle.kts
rnc: Yes because we need to say where to deploy maven central artifacts to (See http://www.albertgao.xyz/2018/01/18/how-to-publish-artifact-to-maven-central-via-gradle/ )

gradle.properties
rnc: I didn't make a change here (apart from trailing newline) - it was already 0.1-SNAPSHOT

build.gradle.kts
geoand: :+1

gradle.properties
geoand: No problem, I got mixed up looking at the github UI

PR #81, opened 2019-06-04
Make application of gme.gradle the first that happens in build.gradle

geoand:
We do this in order to apply the plugin early. This is needed in order
to overcome issues with zip tasks that read the are configured
very early and use the project version


geoand:
Closing in favor of #82

PR #90, opened 2019-06-06
Fix Publishing extension configuration

TomasHofman:


geoand:
I don't have an opinion on this, but it looks reasonable :)

PR #97, opened 2019-06-10
NCL-4692 Handle version ranges during alignment

geoand:
A few things need to come into to play for dynamic versions to work:

* Dependencies with dynamic versions **always** need to be recorded into `manipulation.json`, whether or not the actual dependency was aligned from DA.
* When recording the dependency in `manipulation.json`, the key of the json entry (which is the usual `$groupId:artifactId:$version` format), needs to use the version that is present in the build file (for example `latest.release`, or `2.+`) **instead** of the resolved version. This is done so when the manipulation plugin runs, it can properly look up the dynamic dependency and use the correct version (thus ensuring that the manipulation plugin needs no changes).
* When sending requests to DA, we need to use the version **without** the suffix
* When determining the version so send to DA, we use one of the following:
  - If this is the first time the alignment has been run (hence no `manipulation.json` exists), then we just let gradle resolve the dynamic version and use it to send the request to DA
  - If there is a `manipulation.json` file, then we need to use that


analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentService.java
```diff
@@ -101,7 +122,7 @@ default int order() {
 
         ResponseCustomizer NOOP = new ResponseCustomizer() {
             @Override
-            public Response customize(Response response) {
+            public Response customize(Response response, Request request) {
```
rnc: I'm really not convinced this is needed. I think we need to take another look at how to order and lay this out

geoand: What do you have in mind? Why does this not look an appropriate solution my I ask?

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
```diff
@@ -337,4 +346,95 @@ private void writeRepositorySettingsFile(Collection<ArtifactRepository> reposito
...
+            Set<ProjectRef> dynamicDependencies, Map<ProjectRef, String> dynamicDepToKey) {
+        configuration.getIncoming().getDependencies().withType(ExternalDependency.class).matching(d -> {
+            final String requiredVersion = d.getVersionConstraint().getRequiredVersion();
+            return requiredVersion.endsWith("latest.release") || requiredVersion.endsWith("+");
+        }).forEach(d -> {
+            final SimpleProjectRef ref = new SimpleProjectRef(d.getGroup(), d.getName());
+            dynamicDependencies.add(ref);
+            dynamicDepToKey.put(ref,
+                    new SimpleProjectVersionRef(ref, d.getVersionConstraint().getRequiredVersion()).toString());
+        });
```
rnc: I think we can use the DefaultVersionSelector which should give a much better determination of whether something is dynamic or not. Further I think we need to maintain a mapping between original dep (e.g. 1.1+) and what is passed to DA

geoand: Now that the tests are in place, I can add a wrapper around ProjectVersionRef that includes the original version

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
```diff
@@ -337,4 +346,95 @@ private void writeRepositorySettingsFile(Collection<ArtifactRepository> reposito
...
+    }
+
+    private Set<ProjectVersionRef> updateDynamicDependenciesToExistingVersions(Set<ProjectVersionRef> allDependencies,
+            Set<ProjectRef> dynamicDependencies, Project project) {
```
rnc: Is this to try and pass existing redhat dependency versions to DA for realignment? I am wondering if we should try and split this up (in the sense of multiple PRs) as this is covering too much

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentService.java
rnc: Oh while it does look like it would work I just wonder whether there is a simpler option - for instance maintaining an extra (non-serialised) map in the model which temporarily is used to store the originalKey->resolvedKey (which is passed to DA) .... essentially I'm just thinking about whether we keep things simpler :-)

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
geoand: This is done in order to retrieve the previously aligned version, otherwise repeat builds won't work.

I up to splitting it however you like .

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
rnc: See my above comment - I want to see whether its feasible to take another look and see if its possible to make things simpler rather than more wrappers around things.

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentService.java
geoand: You are probably right! I'll take a look at it with a fresh mind tomorrow 😉

rnc: Cool, will have a think as well, I think we've made good progress establishing a baseline of what is needed.

geoand: Agreed 😎

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
rnc: One point regarding this, while its cleaner (!) than my pseudo code earlier I notice its before we handle unresolved dependencies - I wonder if that could be an issue?

geoand: The actual resolution goes on later on, this part should just be book keeping for later on.
Of course there might be something I have missed, I wouldn't be surprised when it comes to Gradle 😆

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
rnc: A repeat build should work due to the lock file - but a repeat alignment run is another question? I guess the question is whether rerunning alignment, should reuse previously locked dynamic versions or not ? (I don't think concrete versions are an issue) ?

geoand: I would assume that a repeat build should use the same upstream version, but attempt to find a new aligned version for that dependency, correct?
That is what currently happens, and according to the tests, it works (the Undertow version is bumbed to redhat-00002)

geoand:
Closing in favor of #98

PR #98, opened 2019-06-11
 NCL-4692 Handle version ranges during alignment 

rnc:
This combines @geoand work with further changes from myself to see if we can simplify it further.

This is draft as there are still 'TODO:' in the code to discuss.


rnc:
Cross-reference - this layers on top of #98

PR #99, opened 2019-06-12
NCL-4692 Add support for reading version from lock files

geoand:


analyzer/src/main/java/org/jboss/gm/analyzer/alignment/LockfileIO.java
```diff
@@ -0,0 +1,73 @@
+package org.jboss.gm.analyzer.alignment;
```
rnc: I wonder if this should be in common/ .... / io for consistency ?

geoand: I was thinking of it, but since it won't be used in the other module/plugin, I opted to keep it simple for the time being

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
```diff
@@ -269,9 +280,17 @@ private void writeGmeConfigMarkerFile() throws IOException {
                                 dep.toString(), configuration.getName());
                         return;
                     }
+                    String version = dep.getModuleVersion(); // this is the resolved version from gradle
+                    // if the dependency is present in any of the lockfiles, then we use that version
+                    for (ProjectVersionRef lockFileDep : lockFileDeps) {
+                        if (lockFileDep.asProjectRef().getGroupId().equals(dep.getModuleGroup())
+                                && lockFileDep.asProjectRef().getArtifactId().equals(dep.getModuleName())) {
```
rnc: I don't think we need to convert to `asProjectRef` just to extract the groupId/artifactId.

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
```diff
@@ -284,8 +303,6 @@ private void writeGmeConfigMarkerFile() throws IOException {
                     }
 
                     RelaxedProjectVersionRef relaxedProjectVersionRef;
-                    // If we haven't found any original dependency we'll default to the current resolved dependency
-                    // value. This might be possible if the dependency has come from a lock file.
```
rnc: I think this comment is still valid?

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
geoand: Good catch

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
geoand: Yeah, you are probably right. I removed when I was initially thinking of making some change here but ended up not doing some. I'll add it back

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/LockfileIO.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    private static Set<ProjectVersionRef> readProjectVersionRefLocksOfFile(File lockfile) {
+        try {
+            return FileUtils.readLines(lockfile, Charset.defaultCharset())
+                    .stream()
+                    .filter(l -> !l.startsWith("#"))
+                    .map(l -> l.split(":"))
+                    .filter(p -> p.length >= 3)
+                    .map(p -> new SimpleProjectVersionRef(p[0], p[1], p[2]))
```
rnc: I suppose there is not much difference from using a Properties and reader versus the readLines. 
As `SimpleProjectVersionRef.parse(String s)` exists do we need to use the map/filter ?

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
geoand: Fixed

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
geoand: Fixed

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/LockfileIO.java
geoand: Didn't know about that one, will check :)

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/LockfileIO.java
rnc: Yeah see what you're saying. I think I would still prefer it there for consistency (as that packaging is becoming a bit of a 'dumping ground' ... either that or we create a new io subpackage ?

geoand: OK, I'll add it to a new io subpackage :+1:

geoand: But we can of course move it to common if you prefer, up to you!

PR #105, opened 2019-06-13
NCL-4818 Alter every extra gradle file with buildscript to include repos

geoand:
We need this because the buildscript itself can be downloading repos
that we need to proxy through Indy

### DIFF
--- a/analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
+++ b/analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
@@ -213,7 +213,7 @@ public class AlignmentTask extends DefaultTask {
                 StandardCopyOption.REPLACE_EXISTING);
     }
 
-    private void updateAllExtraGradleFilesWithGmeRepos() throws IOException {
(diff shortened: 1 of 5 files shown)


analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
```diff
@@ -198,6 +205,33 @@ private void writeGmeMarkerFile() throws IOException, ManipulationException {
...
+        final File rootDir = getProject().getRootDir();
+        final File gradleScriptsDirectory = rootDir.toPath().resolve("gradle").toFile();
+        if (!gradleScriptsDirectory.exists()) {
+            return;
+        }
+        final Collection<File> extraGradleScripts = FileUtils.listFiles(gradleScriptsDirectory, new SuffixFileFilter(".gradle"),
+                DirectoryFileFilter.DIRECTORY);
+        for (File extraGradleScript : extraGradleScripts) {
+            final List<String> lines = FileUtils.readLines(extraGradleScript, Charset.defaultCharset());
+
```
rnc: Don't we need to use `String first = org.jboss.gm.common.utils.FileUtils.getFirstLine(lines); ` so we don't end up with duplicate lines?

PR #108, opened 2019-06-13
Ensure the projects publishing repos aren't used

geoand:


rnc:
Looks great! Only slight concern (for future) is we don't have a test case and its possible `MavenPublicationRepositoryAction` may suffer the same issue. I vote to merge now though in order to test in pnc.

geoand:
I agree! There could certainly be more issues we run into out along the way :)

PR #111, opened 2019-06-15
NCL-4593 Provide ability to execute a groovy script when done aligning

geoand:


rnc:
Heh, I had very similar thoughts, doing something similar to https://github.com/release-engineering/pom-manipulation-ext/blob/master/core/src/main/java/org/commonjava/maven/ext/core/impl/BaseGroovyManipulator.java#L119

analyzer/src/functTest/java/org/jboss/gm/analyzer/alignment/SimpleProjectWithCustomGroovyScriptFunctionalTest.java
```diff
@@ -0,0 +1,95 @@
...
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.contrib.java.lang.system.RestoreSystemProperties;
+import org.junit.contrib.java.lang.system.SystemOutRule;
+import org.junit.rules.TemporaryFolder;
+import org.junit.rules.TestRule;
+import org.junit.runner.RunWith;
+
+@RunWith(BMUnitRunner.class)
```
rnc: I don't think we need to use Byteman runner if there are no bmrules?

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
```diff
@@ -527,4 +531,20 @@ private String writeProjectNameIfNeeded() throws IOException {
 
         return newProjectName;
     }
+
+    // for now we simply assume that the script is called gme.groovy and already resides in the project's root directory
+    // TODO allow downloading scripts from a configured URL
```
rnc: Agreed, lets get it working then reuse PME FileIO code? ( https://github.com/release-engineering/pom-manipulation-ext/blob/master/core/src/main/java/org/commonjava/maven/ext/core/impl/BaseGroovyManipulator.java#L94 )

analyzer/src/functTest/java/org/jboss/gm/analyzer/alignment/SimpleProjectWithCustomGroovyScriptFunctionalTest.java
geoand: Certainly, I guess it got left over from copy pasta

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
geoand: Makes perfect sense!

analyzer/src/functTest/java/org/jboss/gm/analyzer/alignment/SimpleProjectWithCustomGroovyScriptFunctionalTest.java
geoand: Fixed

rnc:
I think (again for future) we could consider doing something like https://mrhaki.blogspot.com/2013/11/groovy-goodness-set-script-class-with.html (similar to PME) or possibly a DelegatingScript - to give better stronger typing in the groovy.

geoand:
Agreed! We can (and should) make plenty of improvements to the Groovy support in the future.
My only hope is that a limited number of projects will actually need groovy support in the first place 😉

rnc:
Ok to merge - I think we should discuss internally a potential script using this to verify it works.

geoand:
OK, I'll merge for the time being and let's see on Monday how to move forward

PR #116, opened 2019-06-17
Add groovy docs

rnc:


guide/groovy.md
```diff
@@ -0,0 +1,74 @@
...
+
+```
+import groovy.transform.BaseScript
+import org.jboss.gm.analyzer.alignment.groovy.GMEBaseScript
+
+@BaseScript GMEBaseScript gmeScript
+
+```
+
+The scripts will be run **after** the initial alignment.
```
geoand: Perhaps we should say there that the scripts are run as the *last* part of the alignment process?

rnc: Ok sure

PR #126, opened 2019-06-19
Fix broken tests

geoand:
Apply the `gmeAnalyse` property to the classpath in `gme.gradle` as well
to ensure that the manipulation plugin is not looked up when not needed

### DIFF
--- a/analyzer/src/main/resources/gme.gradle
+++ b/analyzer/src/main/resources/gme.gradle
@@ -25,7 +25,7 @@ buildscript {
     // this flag is used to disable the execution of the manipulation plugin
     // after the analyzer plugin has already run on the projects
     // This avoids issues with rerunning in local environments and also in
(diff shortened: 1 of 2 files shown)


analyzer/src/main/resources/gme.gradle
```diff
@@ -22,14 +22,19 @@ buildscript {
         }
     }
 
-    dependencies {
-        classpath "org.jboss.gm.manipulation:manipulation:${project.version}"
+    // this flag is used to disable the execution of the manipulation plugin
+    // after the analyzer plugin has already run on the projects
+    // This avoids issues with rerunning in local environments and also in
+    // tests when we don't want to apply the manipilation plugin, but
```
rnc:
```suggestion
    // tests when we don't want to apply the manipulation plugin, but
```

PR #128, opened 2019-06-21
NCL-4780 Handle non-existing manipulation.json

rnc:


manipulation/src/functTest/java/org/jboss/gm/manipulation/SimpleProjectFunctionalTest.java
```diff
@@ -27,6 +27,25 @@
     @Rule
     public TemporaryFolder tempDir = new TemporaryFolder();
 
+    @Test
+    public void nomanipignore() throws IOException, URISyntaxException {
```
geoand: Can we perhaps improve the name of this test method?

PR #132, opened 2019-08-17
Ncl4672

rnc:
@geoand (WIP to discuss after PTO)

Within this PR I have also refactored the layout slightly and adds a nicer output for the tests.

Its main point was to quieten down the logging so that in PNC its easier to read the alignment log. Things to discuss:
- Should we use Gradle logger ? (https://docs.gradle.org/current/userguide/logging.html)
- Should we output using lifecycle (dependent on above) or info etc.
   - Note that if we need debug logging from the analyser it gets _very_ messy due to the amount pushed out by Gradle - i.e. maybe we should use lifecycle & info. 
- Or potentially create out own wrapper which sends info to llifecycle and debug to info....

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,3 +1,4 @@
+import com.adarshr.gradle.testlogger.theme.ThemeType
 import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
 import java.text.SimpleDateFormat
 import java.util.*
@@ -89,7 +90,7 @@ subprojects {
         // it tries to shadow the entire gradle api
         // Moreover, the gradleApi and groovy dependencies in the plugins themselves
         // have been explicitly declared with the shadow configuration
-        configurations.get("compile").dependencies.remove(dependencies.gradleApi())
+        configurations["compile"].dependencies.remove(dependencies.gradleApi())
 
         // make build task depend on shadowJar
         val build: DefaultTask by tasks
@@ -129,10 +130,10 @@ subprojects {
                     // we publish the init gradle file to make it easy for tools that use
                     // the plugin to set it up without having to create their own init gradle file
                     if (project.name == "analyzer") {
-                        artifact("${sourceSets.main.get().output.resourcesDir}/analyzer-init.gradle", {
+                        artifact("${sourceSets.main.get().output.resourcesDir}/analyzer-init.gradle") {
                             classifier = "init"
                             extension = "gradle"
-                        })
+                        }
                     }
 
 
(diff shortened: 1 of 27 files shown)


geoand:
Cool, I'll check it out in the next few days!

rnc:
@geoand  Cool :-) I'm leaning towards the first **and** third option right now but we can bounce ideas around when we're both back off holiday. 

I also found https://github.com/wfhartford/gradle-dependency-analyze which seems like a nice equivalent of https://github.com/ImmobilienScout24/illegal-transitive-dependency-check (which I use in PME) and reveals how we are using slf4j-api without declaring it.

geoand:
@rnc looks cool! Let's discuss this more when you are back

rnc:
@geoand  WIP of (1) and (2) together with some colours, better context (and the context ability may be essential for other upcoming PNC requirements).

geoand:
Awesome! Let's discuss once you get back

rnc:
Sure :-) I think its almost ready for a proper review now (!) 

Also , tis hopefully may make it easier to add NCL-5068.

geoand:
@rnc should I start looking at the PR now?

build.gradle.kts
```diff
@@ -74,6 +75,7 @@ subprojects {
         apply(plugin = "java-gradle-plugin")
         apply(plugin = "com.github.johnrengelman.shadow")
         apply(plugin = "com.gradle.plugin-publish")
+        apply(plugin = "com.adarshr.test-logger")
```
geoand: The test output looks awesome with this, +100

geoand:
I am seeing a test failure when running locally (`./gradlew clean build`).

```
  org.jboss.gm.analyzer.alignment.AlignmentTaskEmptyVersionTest

    ✘ testProject

      java.lang.reflect.InvocationTargetException
          at org.jboss.gm.analyzer.alignment.AlignmentTaskEmptyVersionTest.testProject(AlignmentTaskEmptyVersionTest.java:96)
      Caused by: org.commonjava.maven.ext.common.ManipulationUncheckedException: For configuration default, unable to resolve all project dependencies: [org.apache.commons:dummy-artifact:]
          at org.jboss.gm.analyzer.alignment.AlignmentTaskEmptyVersionTest.testProject(AlignmentTaskEmptyVersionTest.java:96)

```

When I run from `master` instead of the PR everything is fine. Is it working for you @rnc ?

rnc:
Pushed a fix ( I think!)

geoand:
Tests now pass locally with the latest changes, :+1:

PR #136, opened 2019-09-05
Ncl 5127

geoand:


rnc:
LGTM!

geoand:
Let me make an addition to one of the tests before we merge

analyzer/src/functTest/java/org/jboss/gm/analyzer/alignment/TestUtils.java
```diff
@@ -24,34 +28,62 @@
...
+    static TestManipulationModel align(File projectRoot, boolean expectFailure, Map<String, String> systemProps) {
         assertThat(projectRoot.toPath().resolve("build.gradle")).exists();
 
         final BuildResult buildResult;
         final TaskOutcome outcome;
 
+        final Map<String, String> finalSystemProps = new HashMap<>(systemProps);
+        if (!finalSystemProps.containsKey("repoRemovalBackup")) {
+            finalSystemProps.put("repoRemovalBackup", "settings.xml");
+        }
```
rnc: I don't think this check is needed? Won't later arguments override earlier?

geoand: The idea was to add `-DrepoRemovalBackup=settings.xml` as a default value. If users don't supply a value for `repoRemovalBackup` then `settings.xml` should be used, otherwise their value take precedence.

rnc: Their value would take precedence anyway as its later in the -D list ?

geoand: Aha, now I see what you are saying

PR #146, opened 2019-09-18
NCL-5099 Fix for publishing when module has no java component

TomasHofman:
Another for openjfx.

### DIFF
--- a/manipulation/src/main/java/org/jboss/gm/manipulation/actions/PublishingArtifactsAction.java
+++ b/manipulation/src/main/java/org/jboss/gm/manipulation/actions/PublishingArtifactsAction.java
@@ -2,9 +2,12 @@ package org.jboss.gm.manipulation.actions;
 
 import org.gradle.api.Action;
 import org.gradle.api.Project;
+import org.gradle.api.component.SoftwareComponent;
+import org.gradle.api.logging.Logger;
 import org.gradle.api.publish.PublicationContainer;
 import org.gradle.api.publish.PublishingExtension;
 import org.gradle.api.publish.maven.MavenPublication;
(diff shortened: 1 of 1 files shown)


manipulation/src/main/java/org/jboss/gm/manipulation/actions/PublishingArtifactsAction.java
```diff
@@ -28,22 +29,27 @@
...
+                                    mavenPublication.from(javaComponent);
+                                }
+                            });
+                        }
+                    });
+                }
+            });
+        } else {
+            project.getLogger().warn("No java component found for project {}, no artifact to publish.",
+                    project.getName());
```
rnc: Does this mean we don't publish then ? i.e. the build will complete but nothing will get uploaded?

TomasHofman: For given module, yes. I think root module would be often time the case.

rnc: @TomasHofman Sorry didn't quite understand your last comment? I suppose the problem is at this point we know the build might fail to deploy, but failing in the alignment stage doesn't seem right either.

Potentially switch to error logging to try and highlight it better?

Also please note https://github.com/project-ncl/gradle-manipulator/blob/master/CONTRIBUTING.md#code-recommendations - any logging **must** use GMLogger now.

rnc: Discussed offline and agreed to just change logger and nothing else

TomasHofman:
PR updated.

PR #147, opened 2019-09-22
NCL-4779 Handle Ivy repositories with no URL

rnc:


thescouser89:
Jesus that's a fast fix! Thanks Nick!!!!

PR #156, opened 2019-10-02
NCL-5217 Add test jar publishing for analyzer module

rnc:


rnc:
**( don't merge this yet  ... it might not be needed after all )**

PR #162, opened 2019-10-16
Add Lombok

rnc:
@geoand  Can you try building on command line and importing into IntelliJ to make sure it works for you please.


rnc:
retest this please

PR #177, opened 2019-11-08
Generating correct pom.xml when spring-d-m is used

TomasHofman:
"WebAuthn4J gradle build in PNC does not generate pom.xml files with versions of used dependencies"

NCL-5208

Fixed issues:
* Analyzer: dependencies that inherit version from BOM were missing version in generated manipulation.json keys.
  E.g., manipulation.json would contain:
  ```
      "alignedDependencies" : {
        "ch.qos.logback:logback-classic" : {
          "groupId" : "ch.qos.logback",
          "artifactId" : "logback-classic",
          "version" : "1.2.3.redhat-00001"
        }
      }
  ```
  while it should have been:
  ```
  "alignedDependencies" : {
        "ch.qos.logback:logback-classic:1.2.3" : {
          "groupId" : "ch.qos.logback",
          "artifactId" : "logback-classic",
          "version" : "1.2.3.redhat-00001"
        }
      }
  ```
* Manipulator:
    * Don't prevent spring-d-m from modifying the pom.xml - thus spring-d-m can include BOM references in pom.xml, which resolves the problem with missing dependency versions.
    * Fix version resolution when spring-d-m is used in combination with maven-publish plugin.


TomasHofman:
Please do not merge this yet, I'm gonna test on more real life projects, but I wanted to make this visible. So far I was only testing on webauthn4j.

Interestingly I didn't notice any negative behavior when spring-d-m is allowed to modify the pom.xml. I didn't see it rewrite any of our changes, it just adds managed dependency section with imported BOMs.

rnc:
Thats great!! Thanks very much for this :-)

geoand:
Awesome!

TomasHofman:
I thinks it's ready for review.

Note that I'm not manipulating versions in <managedDependencies> section, only in direct dependencies. So the managed dep will have original version, but that will be effectively rewritten by direct dep which would have updated version.

geoand:
Neat!!!

manipulation/src/functTest/java/org/jboss/gm/manipulation/SimpleProjectWithMisconfiguredSpringDMAndMavenPluginsFunctionalTest.java
```diff
@@ -23,6 +24,7 @@
     public TemporaryFolder tempDir = new TemporaryFolder();
 
     @Test
+    @Ignore("Disabling generatedPomCustomization is no longer necessary.")
     public void ensureProperPomGenerated() throws IOException, URISyntaxException {
```
rnc: If this test is no longer applicable why not just remove it ?

manipulation/src/functTest/java/org/jboss/gm/manipulation/SimpleProjectWithSpringDMAndMavenPluginsFunctionalTest.java
```diff
@@ -48,17 +50,27 @@ public void ensureProperPomGenerated() throws IOException, URISyntaxException, X
                 .forwardOutput()
                 .withPluginClasspath()
                 .build();
-        assertThat(buildResult.task(":install").getOutcome()).isEqualTo(TaskOutcome.SUCCESS);
+        assertThat(Objects.requireNonNull(buildResult.task(":install")).getOutcome()).isEqualTo(TaskOutcome.SUCCESS);
```
rnc: Is this due to warnings in IntelliJ ? - that the code there may throw `NullPointerException`? However, Objects.requireNonNull just throws NullPointerException so I'm not sure there is much benefit gained.

rnc:
This is great! Just a couple of minor questions/comments...

manipulation/src/functTest/java/org/jboss/gm/manipulation/SimpleProjectWithSpringDMAndMavenPluginsFunctionalTest.java
TomasHofman: Yes, this was just to get rid of Intellij warnings. I should probably use assertion instead, will fix.

rnc:
Fantastic work :-)

PR #182, opened 2019-11-19
Improve logging. Remove debug output

rnc:


rnc:
@geoand I think I'll change the groupId - should be transparent to the users but it looks better.

rnc:
( Spoke offline regarding groupId change )

PR #190, opened 2020-01-23
NCL5385 Attempt to workaround archivebasename.

rnc:


manipulation/src/functTest/java/org/jboss/gm/manipulation/SimpleProjectWithMavenPluginBasenameFunctionalTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+        final ManipulationModel alignment = ManipulationIO.readManipulationModel(simpleProjectRoot);
+
+        final BuildResult buildResult = TestUtils.createGradleRunner()
+                .withProjectDir(simpleProjectRoot)
+                //.withDebug(true)
+                .withArguments("uploadArchives", "--stacktrace", "--info")
+                .withPluginClasspath()
+                .forwardOutput()
+                .forwardOutput()
```
TomasHofman: Duplicate line, though no harm...

manipulation/src/main/java/org/jboss/gm/manipulation/actions/LegacyMavenPublishingRepositoryAction.java
```diff
@@ -109,11 +113,34 @@ public void execute(Project project) {
...
+            }
+        });
+        uploadArchives.doLast(action -> {
+            // TODO: Find a better method
+            // Now revert the action performed above.
+            if (abn != null) {
+                logger.warn("Resetting project name after archivesBaseName override to {} from {}", originalName, abn);
+                updateNameField(project, abn);
+            }
+        });
```
TomasHofman: Nice!

manipulation/src/main/java/org/jboss/gm/manipulation/actions/LegacyMavenPublishingRepositoryAction.java
```diff
@@ -122,4 +149,12 @@ public void execute(Project project) {
         install.setConfiguration(installArchives);
         uploadArchives.setConfiguration(publishArchives);
     }
+
+    private void updateNameField(Project project, Object replacement) {
+        try {
+            FieldUtils.writeField(project, "name", replacement, true);
```
TomasHofman: So this has to be reflection, because the Gradle API only allows to change the project name in settings.gradle, which is processed in the earliest stage of the build...

But will this work in JDK 9+? I would expect getting one of those "Unable to make xxx accessible: module X does not 'opens package' to Y"? On top of that "name" is a final field. Who knows what can go wrong :).

But I guess we won't need to build in JDK 11 for some time yet...
Hope nobody ever see this :).

rnc: Yes exactly. Commons lang does support it https://commons.apache.org/proper/commons-lang/changes-report.html in general but if we get problems we may have to reinvestigate this or suggest the project resolves it outside of the tool.

TomasHofman: +1

PR #218, opened 2020-07-25
NCL5499 Add testing under Gadle 6

rnc:


rnc:
@geoand @dwalluck So, adding official Gradle 6.x compatibility hit a problem - 6.3,4 all work fine (and I added a 6.4 test as can be seen in travis) but 6.5 fails with an a very bizarre InvalidJarIndexException from deep inside Gradle ( https://gist.github.com/rnc/5d62bb5565f4db8636501977dbcc0724 ). Any ideas welcome! :-)

geoand:
Wow.... That looks nasty! Perhaps it's a Gradle bug? Have you tried the rc builds of 6.6?

rnc:
@geoand No, I will try that next. We already hit a gradle bug in the alignmenttask code - see https://github.com/project-ncl/gradle-manipulator/blob/master/analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java#L394 :-(

geoand:
Yeah I am not surpsised... Gradle moves very fast so bugs are inevitable :)

rnc:
And ... nope ... 6.6 rc3 doesn't work either :-(

geoand:
Darn... Since the exception is so weird, maybe it makes sense to contact the Gradle team? I don't really see what can be done on this end :(

rnc:
Yes may have to try that and also whether its something we can debug into to understand. Mind, I think the gradle team are a touch overloaded .... currently over 2300 issues open!

geoand:
Ouch....

dwalluck:
Maybe there is actually a jar with a bad index causing this exception?

rnc:
@dwalluck Its on every functional test, and its in the junit tmp directories so I suspect not.

rnc:
@geoand @dwalluck To move this forward I have commented out the travis build for 6.5.1 and created a separate ticket internally for that.

PR #219, opened 2020-08-13
NCLSUP-101: Detect when unsupported Gradle or Java versions are used

dwalluck:


rnc:
Nice :-) I've tried it out with Java 11 on an early Gradle and get a rather large stack trace e.g. https://gist.github.com/rnc/b0fc6711cb15aaa3977d19e437196ab5 ; I'm not convinced that the error handling isn't lost in the 'noise' - what do you think?

Another point is that I suspect that it will have to 'wait' to download the appropriate Gradle version before validating the wrapper - while this doesn't add a _lot_ of time it is something to consider.

dwalluck:
> Nice :-) I've tried it out with Java 11 on an early Gradle and get a rather large stack trace e.g. https://gist.github.com/rnc/b0fc6711cb15aaa3977d19e437196ab5 ; I'm not convinced that the error handling isn't lost in the 'noise' - what do you think?

If just the noise is the issue, one option is to remove the exception for the error log message and place it in the debug log message. It was also for potential handling of the error as well as printing a more specific message to the user. For the end user, it may make sense to hide the exception anyway.

> Another point is that I suspect that it will have to 'wait' to download the appropriate Gradle version before validating the wrapper - while this doesn't add a _lot_ of time it is something to consider.

There are two possibilities that I see: check if there is still a way to handle the gradle version using the Gradle API, but without triggering that exception, or, try parsing the version ourselves using a regex. Note that we have at least two paths to a distribution (gradle wrapper and user supplied path via the `-l` option). I suppose with the gradle wrapper it may be safe to assume the URL format (if using official upstream URLs at least). With the user supplied option, it will depend on the path, and therefore, both OS and OS distribution, so this isn't a great candidate for parsing, but we could still try to match gradle-<version> in the path.

rnc:
> > Nice :-) I've tried it out with Java 11 on an early Gradle and get a rather large stack trace e.g. [gist.github.com/rnc/b0fc6711cb15aaa3977d19e437196ab5](https://gist.github.com/rnc/b0fc6711cb15aaa3977d19e437196ab5) ; I'm not convinced that the error handling isn't lost in the 'noise' - what do you think?
> 
> If just the noise is the issue, one option is to remove the exception for the error log message and place it in the debug log message. It was also for potential handling of the error as well as printing a more specific message to the user. For the end user, it may make sense to hide the exception anyway.
>

Yes I agree :-) Something like
```
                        if (m.matches()) {
                            String v = m.group("version");
                            String msg = String.format("Java %s is incompatible with Gradle version used to build", v);

                            logger.debug("Caught exception processing Gradle API - " + msg, e);
                            throw new ManipulationException(msg);
                        }
 ```
I'll comment inline. 

> > Another point is that I suspect that it will have to 'wait' to download the appropriate Gradle version before validating the wrapper - while this doesn't add a _lot_ of time it is something to consider.
> 
> There are two possibilities that I see: check if there is still a way to handle the gradle version using the Gradle API, but without triggering that exception, or, try parsing the version ourselves using a regex. Note that we have at least two paths to a distribution (gradle wrapper and user supplied path via the `-l` option). I suppose with the gradle wrapper it may be safe to assume the URL format (if using official upstream URLs at least). With the user supplied option, it will depend on the path, and therefore, both OS and OS distribution, so this isn't a great candidate for parsing, but we could still try to match gradle- in the path.


Agreed. Perhaps make a short note of these comments on NCLSUP101 for posterity and leave it as you have it as I think its a definite improvement.

File: cli/src/main/java/org/jboss/gm/cli/Main.java [hunk 1]
```diff
@@ -111,8 +119,62 @@ public int run(String[] args) throws Exception {
         return result;
     }
 
-    private void executeGradle() throws ManipulationException {
+    private void logAndThrow(String message, Throwable cause) throws ManipulationException {
```
rnc, line 122: If we a simplifying as per the other comments I don't think we need this indirection

cli/src/main/java/org/jboss/gm/cli/Main.java
```diff
@@ -111,8 +119,62 @@ public int run(String[] args) throws Exception {
...
+            GradleVersion version = GradleVersion.version(versionString);
+
+            logger.info("Gradle version: {}", versionString);
+            logger.info("Java home: {}", env.getJava().getJavaHome());
+            logger.info("JVM arguments: {}", env.getJava().getJvmArguments());
+
+            if (version.compareTo(MIN_GRADLE_VERSION) < 0) {
+                String msg = String.format("%s is too old and is unsupported. You need at least %s.", version,
+                        MIN_GRADLE_VERSION);
+                logAndThrow(msg, null);
```
rnc, line 145: So this could be
```
                throw new ManipulationException("{} is too old and is unsupported. You need at least {}.", version,
                        MIN_GRADLE_VERSION););
```

rnc:
@dwalluck Can we look at this again please to finish this off - I think there are some outstanding comments.

cli/src/main/java/org/jboss/gm/cli/Main.java
dwalluck: I've removed this method.

cli/src/main/java/org/jboss/gm/cli/Main.java
dwalluck: I've changed all of the `throw new ManipulationException` to use this format.

rnc (approved):

PR #220, opened 2020-09-10
NCLSUP116: Handle shaded jar and resolution strategies

rnc:


rnc:
cc @dwalluck

dwalluck:
Looks good, but it will purposely fail by default?

I am having trouble seeing what setting actually works for the most general case for the end user. If it isn't set to `true`, then isn't it more likely to cause a problem with mixing transitive dependencies with overridden dependencies?

If `true` doesn't work in the general case, then it's somewhat unclear what the user should do. Suppose a user gets this message, then picks a value (`true` or `false`), then tries to build and it fails again, then are they likely to just open another ticket?

rnc:
> Looks good, but it will purposely fail by default?
> 
> I am having trouble seeing what setting actually works for the most general case for the end user. If it isn't set to `true`, then isn't it more likely to cause a problem with mixing transitive dependencies with overridden dependencies?
> 
> If `true` doesn't work in the general case, then it's somewhat unclear what the user should do. Suppose a user gets this message, then picks a value (`true` or `false`), then tries to build and it fails again, then are they likely to just open another ticket?

I am viewing the shaded jar as a rare case and want to maintain aligning only direct dependencies. Historically, when PME aligned transitive there have been issues. Hence, I want to keep the general case, but not silently not align the transitive when the shaded plugin is active _unless_ the user specifically chooses not to. So yes, it may initially fail but I want the user to make an informed choice depending upon their build.

dwalluck:
> I am viewing the shaded jar as a rare case and want to maintain aligning only direct dependencies. Historically, when PME aligned transitive there have been issues. Hence, I want to keep the general case, but not silently not align the transitive when the shaded plugin is active _unless_ the user specifically chooses not to. So yes, it may initially fail but I want the user to make an informed choice depending upon their build.

OK, because that's the other thing I was thinking---if you can detect the usage of shadowjar and enable it. I see the reason for leaving the general case as-is.

PR #235, opened 2020-10-05
NCLSUP153 Improve logging and sort manipulation.json

rnc:


analyzer/src/main/java/org/jboss/gm/analyzer/alignment/WithCustomizersDelegatingAlignmentService.java
```diff
@@ -20,6 +22,8 @@
  */
 public class WithCustomizersDelegatingAlignmentService implements AlignmentService {
 
+    private final Logger logger = GMLogger.getLogger(getClass());
```
dwalluck, line 25: Just a note: this `Logger` you haven't made `static`, but the one in `DependencyExclusionCustomizer` is `static`. I always make my loggers `static`, but I haven't read much about which is considered the best practice for loggers.

rnc: Good point/question ; I think from reading on it there are pros/cons ; static is more common though (although then causes issues with inheritance like with the pluginremoval work).

PR #236, opened 2020-10-07
Add .gitattributes to work around spotless issue with eol

dwalluck:
See <https://github.com/diffplug/spotless/issues/39>.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+* text=auto
+* text eol=lf
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,4 @@
 * text=auto
 * text eol=lf
+*.bat text eol=crlf
+*.jar binary


rnc:
Is spotless still causing an issue for you - I thought you were now on Linux? If this has no affects on Linux I'm fine with this.

rnc (approved):

dwalluck:
> Is spotless still causing an issue for you - I thought you were now on Linux? If this has no affects on Linux I'm fine with this.

This was happening while back. I am not sure I knew that it was spotless that was causing it until I saw the same behavior in another repository. That led me to that ticket saying that for some reason spotless will change all of the newlines and for some reason the only way to configure it seems to be through a `.gitattributes` file which it reads.

dwalluck:
Unrelated to this specific issue, this file helps keep eols from changing accidentally. For example, it looks like the `gradlew.bat` file has an eol of `lf` even though it's a DOS file which would have `crlf`. Is this the only Windows file in the codebase? It looks like someone has accidentally changed the eol on the `.bat` file. I tried "upgrading" the gradle wrapper to the same version, and the new `.bat` file definitely has crlf, but it has also changed `-all` zip to `-bin` zip. Is that OK?

```
diff --git a/gradle/wrapper/gradle-wrapper.properties b/gradle/wrapper/gradle-wrapper.properties
index 4e9a67a..5028f28 100644
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,6 +1,5 @@
-#Mon Apr 29 18:31:13 BST 2019
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.6.4-bin.zip
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-5.6.4-all.zip
```

rnc:
I'd rather keep the wrapper distribution type as all. See https://mrhaki.blogspot.com/2016/09/gradle-goodness-specify-wrapper-version.html


> Unrelated to this specific issue, this file helps keep eols from changing accidentally. For example, it looks like the `gradlew.bat` file has an eol of `lf` even though it's a DOS file which would have `crlf`. Is this the only Windows file in the codebase? It looks like someone has accidentally changed the eol on the `.bat` file. I tried "upgrading" the gradle wrapper to the same version, and the new `.bat` file definitely has crlf, but it has also changed `-all` zip to `-bin` zip. Is that OK?
> 
> ```
> diff --git a/gradle/wrapper/gradle-wrapper.properties b/gradle/wrapper/gradle-wrapper.properties
> index 4e9a67a..5028f28 100644
> --- a/gradle/wrapper/gradle-wrapper.properties
> +++ b/gradle/wrapper/gradle-wrapper.properties
> @@ -1,6 +1,5 @@
> -#Mon Apr 29 18:31:13 BST 2019
>  distributionBase=GRADLE_USER_HOME
>  distributionPath=wrapper/dists
> +distributionUrl=https\://services.gradle.org/distributions/gradle-5.6.4-bin.zip
>  zipStoreBase=GRADLE_USER_HOME
>  zipStorePath=wrapper/dists
> -distributionUrl=https\://services.gradle.org/distributions/gradle-5.6.4-all.zip
> ```

dwalluck:
> I'd rather keep the wrapper distribution type as all. See https://mrhaki.blogspot.com/2016/09/gradle-goodness-specify-wrapper-version.html

I figured out how to set it in the build file too. I will submit it as a separate PR I guess it's only tangentially related.

PR #238, opened 2020-10-07
Add a newline at eof, replace tabs, and remove tailing whitespace

dwalluck:
This a whitespace-only patch to unify formatting in the codebase. In particular, some `.gradle` files and a few `.java` files were using tabs instead of spaces which my editor is just going to replace anyway. Also, some files did not end in a newline.


rnc:
What did you change on the binary jar?

dwalluck:
> What/why did you change on the binary jar?

I forgot to exclude it when editing the files, and it broke the build so that wasn't good! Fixed.

rnc (approved):

PR #266, opened 2021-01-27
NCLSUP-250 Handle lazy configured maven publish plugin

rnc:
The original problem was caused by the work in #242 but due to task avoidance configuration in other projects this caused problems. This change reworks that so the 'sledgehammer' solution is only used for the legacy plugin and an approach using available API is taken the current plugin.


manipulation/src/functTest/resources/mongo-java-driver/gradle/publish.gradle
```diff
@@ -102,7 +102,7 @@ configure(javaProjects) { project ->
 
     afterEvaluate {
         jar configureJarManifestAttributes(project)
-        publishing.publications.mavenJava.artifactId = project.archivesBaseName
+//        publishing.publications.mavenJava.artifactId = project.archivesBaseName
```
dwalluck, line 105: What was this originally doing?

manipulation/src/functTest/java/org/jboss/gm/manipulation/ManifestVerificationFunctionalTest.java
```diff
@@ -186,14 +189,20 @@ public void verifyMongoManifest() throws IOException, URISyntaxException {
...
 
-        assertThat(Objects.requireNonNull(buildResult.task(":bson:publishToMavenLocal")).getOutcome())
+        assertThat(Objects.requireNonNull(buildResult.task(":bson:publish")).getOutcome())
                 .isEqualTo(TaskOutcome.SUCCESS);
 
+        Path pathToArtifacts = publishDirectory.toPath().resolve(PATH_IN_REPOSITORY);
+        assertTrue(systemOutRule.getLog().contains(
+                "Updating publication artifactId (driver-core) as it is not consistent with archivesBaseName (mongodb-driver-core)"));
+        assertThat(pathToArtifacts.resolve(ARTIFACT_NAME + ".pom").toFile().exists()).isTrue();
+        assertThat(pathToArtifacts.resolve(ARTIFACT_NAME + ".jar").toFile().exists()).isTrue();
```
dwalluck, line 204: Can't this just be `assertThat(pathToArtifacts.resolve(ARTIFACT_NAME + ".jar").toFile()).exists();`

manipulation/src/functTest/resources/mongo-java-driver/gradle/publish.gradle
rnc: This was for that project setting the artifactId to correspond to instructions in https://docs.gradle.org/current/userguide/maven_plugin.html#uploading_to_maven_repositories However, I wanted to verify that the changes I made could handle the situation where this _wasn't_ done hence commenting it out.

manipulation/src/functTest/java/org/jboss/gm/manipulation/ManifestVerificationFunctionalTest.java
rnc: Huh, probably. For some reason a lot are probably like that :-(

rnc: Actually, no. Those assertThat are returning an instance of  `BooleanAssert` so it needs an final test (i.e. the `isTrue`, or elesewhere `isFalse` ) otherwise the statement becomes a NOP.

PR #268, opened 2021-03-05
Add check for build JDK

rnc:


dwalluck:
Just to verify, I checked that we already set Java 8 source/target here https://github.com/project-ncl/gradle-manipulator/blob/2137314dc050092591143293548266779b1be479/build.gradle.kts#L301-L304

PR #269, opened 2021-03-11
NCLSUP281 Handle publishing for buildSrc. Workaround partial disablin…

rnc:
…g of generatePomFile tasks. Allow disabling of manipulation plugin


rnc:
cc @dwalluck

rnc:
For future reference : annoyingly codecov is stating the diff didn't test certain lines when tests I added assert output from those lines. All CI passed, even though codecov is reporting failure.

PR #272, opened 2021-03-15
NCL-6468 Add ability to specify custom version of ManipulationPlugin

rnc:


analyzer/src/functTest/java/org/jboss/gm/analyzer/alignment/SimpleProjectFunctionalTest.java
```diff
@@ -101,6 +101,37 @@ public void ensureAlignmentFileCreated() throws IOException, URISyntaxException,
...
+        assertThat(alignmentModel).isNotNull().satisfies(am -> {
+            assertThat(am.getOriginalVersion()).isEqualTo("1.0.1");
+            assertThat(am.getGroup()).isEqualTo("org.acme.gradle");
+            assertThat(am.getName()).isEqualTo("root");
+        });
+        File gme = new File(projectRoot, AlignmentTask.GME);
+        assertTrue(gme.exists());
+        assertThat(FileUtils.readFileToString(gme, Charset.defaultCharset()))
+                .contains("org.jboss.gm:manipulation:0.1.2.3.4");
+        assertTrue(new File(projectRoot, AlignmentTask.GME).exists());
```
dwalluck, line 121: Line 130 has assertThat().exists() but there has assertTrue().

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
```diff
@@ -284,13 +284,23 @@ public void perform() {
...
         }
+        if (!isEmpty(configuration.manipulationVersion())) {
+            String gmeGradleString = FileUtils.readFileToString(gmeGradle, Charset.defaultCharset());
+            String currentVersion = gmeGradleString.replaceFirst(
+                    "(?s).*(classpath \"org.jboss.gm:manipulation:)([0-9]+\\.[0-9]+(-SNAPSHOT)??)\".*", "$2");
+            logger.info("Replacing version {} with {} for the ManipulationPlugin", currentVersion,
+                    configuration.manipulationVersion());
+            FileUtils.writeStringToFile(gmeGradle,
+                    gmeGradleString.replaceFirst(currentVersion, configuration.manipulationVersion()),
+                    Charset.defaultCharset());
```
dwalluck, line 302: I have said before I think it is correct to use `StandardCharsets.UTF_8` as on Windows at least `Charset.defaultCharset()` returns `Cp1252` and I am not well-versed in the differences, at least until this <https://bugs.openjdk.java.net/browse/JDK-8187041> gets accepted.

analyzer/src/functTest/java/org/jboss/gm/analyzer/alignment/SimpleProjectFunctionalTest.java
rnc: Do you mean for consistency or some other reason? I don't see why `assertTrue` shouldn't work?

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
rnc: We use `Charset.defaultCharset` everywhere so I'm not convinced its worth changing? We don't even validate on Windows do we?

analyzer/src/functTest/java/org/jboss/gm/analyzer/alignment/SimpleProjectFunctionalTest.java
dwalluck: Yes, just for consistency of the calls within the same method and also that `assertThat()` will produce a more informative error message should the test ever fail.

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
dwalluck: Windows was just an example, although I thought that at least something was running on Windows, PME I guess.. It doesn't just affect Windows, it affects _all_ operating systems, as macOS, is always UTF-8 except in the POSIX C locale and Linux can always be changed via the locale or lang settings.

Probably since the files we write are generally within normal ASCII range it would not fai

analyzer/src/functTest/java/org/jboss/gm/analyzer/alignment/SimpleProjectFunctionalTest.java
rnc: Actually there is a lot of usage of e.g. `assertTrue(new File(projectRoot, AlignmentTask.GME).exists());` in that file.

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
rnc: Given that almost everywhere is using `defaultCharset` at the moment I'm inclined to leave this as is.

analyzer/src/functTest/java/org/jboss/gm/analyzer/alignment/SimpleProjectFunctionalTest.java
rnc: Regex replace in intellij to the rescue.

PR #276, opened 2021-03-30
Whitespace only test change

rnc:
_do not merge_ ; testing CI (cc @dwalluck )

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -24,6 +24,8 @@ plugins {
     id("org.kordamp.gradle.jacoco") version "0.42.1"
 }
 
+
+
 apply(plugin = "net.researchgate.release")
 
 if(!JavaVersion.current().isJava11Compatible){


rnc:
@dwalluck Looks like something is wrong with gh actions - the ci was working (for code) 2 weeks ago.

dwalluck:
I was checking to see if we cache the gradle cache and it looks like we do, so why would it even be required to download dependencies if nothing has changed?

rnc:
Local error seems to be 
```
      org.gradle.internal.resolve.ModuleVersionResolveException: Could not resolve io.undertow:undertow-core:2.0+.
      Required by:
          project :
      Cause 1: org.gradle.internal.resolve.ModuleVersionResolveException: Could not resolve io.undertow:undertow-core:2.0.30.SP3-redhat-00001.
```

PR #278, opened 2021-04-13
NCL-6081: Generate alignment report

dwalluck:


analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
```diff
@@ -84,6 +91,7 @@
     public static final String APPLY_GME_REPOS = "buildscript { apply from: new File(buildscript.getSourceFile().getParentFile(),\"gme-repos.gradle\"), to: buildscript }";
     public static final String GME_PLUGINCONFIGS = "gme-pluginconfigs.gradle";
     public static final String NAME = "generateAlignmentMetadata";
+    public static final String MANIPULATION_PME_BASENAME = "manipulation-pme";
```
rnc, line 94: Where did you get this name from - I didn't think this was agreed in the ticket?

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
```diff
@@ -105,8 +113,10 @@ public void perform() {
         String groupId = ProjectUtils.getRealGroupId(project);
         String projectName = project.getName();
 
-        if (!configOutput.get().getAndSet(true)) {
-            // Only output the config once to avoid noisy logging.
+        final StringBuilder builder = new StringBuilder(500);
+
+        // Only output the config once to avoid noisy logging.
+        if (logger.isInfoEnabled() && !configOutput.get().getAndSet(true)) {
```
rnc, line 119: Why change this?

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
```diff
@@ -257,10 +266,66 @@ public void perform() {
...
+                        final String group = correspondingModule.getGroup().isEmpty() ? alignmentModel.getGroup()
+                                : correspondingModule.getGroup();
+                        final String artifact = correspondingModule.getName();
+                        final String ga = group + ":" + artifact;
+                        final String originalModuleGav = ga + ":" + (correspondingModule.getOriginalVersion() == null
+                                ? ""
+                                : correspondingModule.getOriginalVersion());
+                        final String newModuleGav = ga + ":" + correspondingModule.getVersion();
+                        logger.debug("Relocation : {} ---> {}", originalModuleGav, newModuleGav);
+                        append(builder, "\tRelocation : {} ---> {}", originalModuleGav, newModuleGav);
```
rnc, line 305: I am seeing project sub-modules being labelled as relocations which does not seem right.

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
```diff
@@ -257,10 +266,66 @@ public void perform() {
...
+                    }
+                }
+
+                final String report = builder.toString();
+                logger.info(report);
+                writeReport(project.getRootDir(), report);
+                writeJsonModel(project.getRootDir(), jsonReport);
                 GroovyUtils.runCustomGroovyScript(logger, InvocationStage.LAST, project.getRootDir(), configuration,
                         project.getRootProject(),
                         alignmentModel);
```
rnc, line 328: What happens if the groovy scripting makes further changes? This order does not seem right.

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
```diff
@@ -644,4 +711,37 @@ private void writeRepositorySettingsFile(Map<ArtifactRepository, Path> repositor
         return allDependencies;
     }
 
+    public void writeReport(File rootDir, String report) throws ManipulationException {
+        final File reportFile = new File(".", "report-pme-"
+                + RandomStringUtils.randomAlphanumeric(10) + ".txt");
```
rnc, line 716: I don't think using `randomAlphanumeric` is correct - we had already agreed a naming format and this makes it unpredictable. We should review where PME puts the files - putting them in a non-build directory may mean repour commits them which might not be desirable.

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
```diff
@@ -644,4 +711,37 @@ private void writeRepositorySettingsFile(Map<ArtifactRepository, Path> repositor
...
+     * Wraps appending to the string builder using SLF4J style substitutions.
+     *
+     * @param builder the string builder
+     * @param message the message (possibly with parameters)
+     * @param args optional parameters.
+     */
+    private static void append(StringBuilder builder, String message, Object... args) {
+        builder.append(MessageFormatter.arrayFormat(message, args).getMessage());
+        builder.append('\n');
+    }
```
rnc, line 746: Nice, although I do wonder if it would be better in common?

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
```diff
@@ -644,4 +711,37 @@ private void writeRepositorySettingsFile(Map<ArtifactRepository, Path> repositor
...
 }
```
rnc, line 747: I think there needs to be tests for these changes.

rnc (changes requested):

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
dwalluck: That was to guard running the code if the logger is not enabled. It is not a simple String, it's a loop for the method called there.

I was having an issue of no test XML file containing the logger output for the alignment task.

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
dwalluck: OK, then this isn't entirely correct. Is it only printing submodules, or submodules are included and shouldn't be?

I guess I had trouble understanding the distinction as all project "dependencies" were also called "modules" or "children".

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
dwalluck: Yes, I just copied this one from PME (more or less).

There is an issue with what separator to use, since normally if we print this to the console we would use `System.lineSeparator()`, but if we want to save it to a text file, it might be better to normalize it to `\n` so that it produces the exact content across platforms (of course, we won't hit this in practice, so it's not that crucial).

If we were able to print it on the fly, we could just call `log`, but I guess PME is collecting it as a string just to be able to have all the content available for a file later?

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
dwalluck: I know it is not correct.

If I use `project.getRootDir()`, I don't find the files at all in the directory tree. I wanted to check what `project.getRootDir()` is being set to, but I cannot see any logging output from the tests.

If I set to `.`, then it adds it in the source tree, which means the files overwrite themselves on the next test and I couldn't see them. This is the reason for adding the random component just for now.

I would have hoped that tests would run outside of the source tree as well. Creating files inside the source tree is not right. For example, maven uses `target/`, then the build files don't pollute the source tree.

The files are always created here, but in PME, the report txt is optional, and doesn't appear to have an official name. It looks like the default is just empty string `""`. I don't really see equivalent options in GME. But, does this file have a standard name?

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
dwalluck: I wasn't sure if it was going to be `alignmentReport.json` or involve renaming `gme.json`. Also, what should the text version of the report be called? I think ` `alignmentReport.txt` makes sense in that case.

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
rnc: Re logger : ok.
Re test xml : I don't follow?

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
rnc: It would help if there were tests (and eventually documentation) I think. Regardless, a submodule version change is not a relocation I would think

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
rnc: There are a lot of separate comments here which probably should be discussed.

I think the output should go in the build directory. But even then, repeated runs, would overwrite (which I would expect).

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
dwalluck: I don't see any of the output on the console, in files, or anywhere. For example: `grep -r "Starting alignment task for project"` returns no matching logs even though that code is obviously running.

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
dwalluck: You are correct. I meant to have these run last, actually.

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
dwalluck: Yes, but it's not really possible to write a test of the output files if the output cannot be confined to a directory specific to the test being run. It's possible to have a unit test with a pre-existing test resource, I suppose, but what about the functional tests?

If there are tests already doing this, I am not sure, it looks like all of the files are confined to the project root, but this is a problem, don't we want to copy all the test resources over before running the tests anyway?

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
```diff
@@ -644,4 +661,124 @@ private void writeRepositorySettingsFile(Map<ArtifactRepository, Path> repositor
         return allDependencies;
     }
 
+    private void writeReport(Path rootDir, String filename, String text) throws ManipulationException {
+        final Path reportFile = Paths.get(filename);
+        final Path reportOutputFile;
+
+        if (reportFile.isAbsolute()) {
```
dwalluck, line 668: We can't default to the project dir ahead of time, I think, so this is what I came up with.

analyzer/src/functTest/java/org/jboss/gm/analyzer/alignment/SimpleProjectFunctionalTest.java
```diff
@@ -211,4 +220,134 @@ public void verifySnapshotHandling() throws IOException, URISyntaxException, Man
...
+        assertThat(jsonFile).isFile();
+        assertThat(textFile).doesNotExist();
+        final PME jsonReport = JSONUtils.fileToJSON(jsonFile);
+        final String jsonString = FileUtils.readFileToString(jsonFile, StandardCharsets.UTF_8);
+        assertThat(jsonString).isEqualTo("{\n" +
+                "  \"executionRoot\" : {\n" +
+                "    \"groupId\" : \"org.acme.gradle\",\n" +
+                "    \"artifactId\" : \"root\",\n" +
+                "    \"version\" : \"1.0.1\"\n" +
+                "  },\n" +
```
dwalluck, line 241: This next gav is appearing twice, since it adds one for the project itself, and then apparently it also appears as a module. Is this correct?

common/src/main/java/org/jboss/gm/common/Configuration.java
```diff
@@ -171,6 +174,18 @@ default Properties getProperties() {
     @DefaultValue("")
     String manipulationVersion();
 
+    @Key("reportTxtOutputFile")
+    @DefaultValue(REPORT_TXT_OUTPUT_FILE)
```
rnc, line 178: Why do we need the field here? Why not just `""` ?

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
```diff
@@ -644,4 +656,124 @@ private void writeRepositorySettingsFile(Map<ArtifactRepository, Path> repositor
...
+
+        final String reportText = builder.toString();
+        logger.info("{}{}", System.lineSeparator(), reportText);
+
+        if (!StringUtils.isEmpty(configuration.reportTxtOutputFile())) {
+            writeReport(rootDir.toPath(), configuration.reportTxtOutputFile(), reportText);
+        }
+
+        if (!StringUtils.isEmpty(configuration.reportJsonOutputFile())) {
+            writeReport(rootDir.toPath(), configuration.reportJsonOutputFile(),
```
rnc, line 775: Shouldn't this output to the build directory not just `rootDir` ? (i.e. as per https://github.com/release-engineering/pom-manipulation-ext/blob/master/core/src/main/java/org/commonjava/maven/ext/core/ManipulationManager.java#L232 )

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
```diff
@@ -644,4 +656,124 @@ private void writeRepositorySettingsFile(Map<ArtifactRepository, Path> repositor
...
+                    } else {
+                        final String originalDependencyGav = String.format("%s:%s:%s", p.getGroupId(), p.getArtifactId(),
+                                p.getVersionString());
+                        final String newDependencyGav = String.format("%s:%s:%s", newDependencyVersion.getGroupId(),
+                                newDependencyVersion.getArtifactId(), newDependencyVersion.getVersionString());
+                        append(builder, "\tDependencies : {} --> {}", originalDependencyGav, newDependencyGav);
+                        dependencies.put(d.toString(), newDependencyVersion);
+                    }
+                });
+            }
```
rnc, line 751: Couple of questions:

1. Is there a reason sub-modules are treated as dependencies? Shouldn't they be separated out as its rather confusing to see a submodule be a dependency?
2. Do you not (currently?) have a capability to separate out dependencies from different sub-modules?

analyzer/src/functTest/java/org/jboss/gm/analyzer/alignment/SimpleProjectFunctionalTest.java
rnc: As per https://github.com/release-engineering/pom-manipulation-ext/blob/master/common/src/main/java/org/commonjava/maven/ext/common/json/PME.java it was to denote where pme was run from and was used by repour. Its probably with syncing with @thescouser89 to check how repour works with gme in this case.

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
```diff
@@ -644,4 +656,124 @@ private void writeRepositorySettingsFile(Map<ArtifactRepository, Path> repositor
...
+        }
+
+        final String reportText = builder.toString();
+        logger.info("{}{}", System.lineSeparator(), reportText);
+
+        if (!StringUtils.isEmpty(configuration.reportTxtOutputFile())) {
+            writeReport(rootDir.toPath(), configuration.reportTxtOutputFile(), reportText);
+        }
+
+        if (!StringUtils.isEmpty(configuration.reportJsonOutputFile())) {
```
rnc, line 774: Can this ever be empty?

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
rnc: By the tie this runs, surely there will be a root project with a build directory?

analyzer/src/functTest/java/org/jboss/gm/analyzer/alignment/SimpleProjectFunctionalTest.java
```diff
@@ -211,4 +220,134 @@ public void verifySnapshotHandling() throws IOException, URISyntaxException, Man
...
+                "------------------- project org.acme.gradle:root:1.0.1",
+                "\tProject version : 1.0.1 --> 1.0.1.redhat-00002",
+                "",
+                "\tDependencies : org.acme.gradle:root:1.0.1 --> org.acme.gradle:root:1.0.1.redhat-00002",
+                "\tDependencies : com.yammer.metrics:metrics-core:2.2.0 --> com.yammer.metrics:metrics-core:2.2.0-redhat-00001",
+                "\tDependencies : org.hibernate:hibernate-core:5.3.7.Final --> org.hibernate:hibernate-core:5.3.7.Final-redhat-00001",
+                "\tDependencies : io.undertow:undertow-core:2.0.15.Final --> io.undertow:undertow-core:2.0.15.Final-redhat-00001",
+                "",
+                "\tNon-Aligned Dependencies : junit:junit:4.12",
+                "\tNon-Aligned Dependencies : org.apache.commons:commons-lang3:3.8.1",
```
rnc, line 350: Can these be sorted alphabetically?

analyzer/src/functTest/java/org/jboss/gm/analyzer/alignment/SimpleProjectFunctionalTest.java
dwalluck: OK, I think the reason it is printed twice here are in the text is that it's treating a module as a dependency?

common/src/main/java/org/jboss/gm/common/Configuration.java
dwalluck: I originally had `""`. I changed it

1. To be consistent with the JSON property.
2. I thought I might have to reference it to set the value from within the tests (but I guess not).
3. It could make sense to make all the keys constants. Otherwise, you have to remember what the key is called and possibly get an error whereas if it was a public constant the IDE would just look it up and ensure you do not mistype or retype these.

If I look at PME both the key and value are constants, which I like:

```
final String reportJsonOutputFile = session.getUserProperties().getProperty( REPORT_JSON_OUTPUT_FILE,
                                                                                             session.getTargetDir() + File.separator + REPORT_JSON_DEFAULT);
``

But, you see that here it is able to get the `targetDir` on the fly, we can't do this because the default is set ahead of time using the `@DefaultValue`, right?

File: analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java (hunk 11 above)
Comment (dwalluck) [reply analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java:775]: I saw GME writing all other files to `rootDir` I guess that's where I got that, but see the above comment about `@DefaultValue`.

It looks like in PME this value is _always_ relative, in which case I should always resolve the property relative to the output directory. Fixing the location of the output is a separate issue.

File: analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java (hunk 12 above)
Comment (dwalluck) [reply analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java:751]: So, it's not really intentional if I am misunderstanding something here where PME is also separating it out. 

I tried to follow what I saw in PME, with this loop being the equivalent of PMEs `compareDependencies( DEPENDENCIES, ...)`, or so I thought.

Because the loop does look up relative to `correspondingModule`, I thought it is separated by module.

File: common/src/main/java/org/jboss/gm/common/Configuration.java (hunk 10 above)
Comment (rnc) [reply common/src/main/java/org/jboss/gm/common/Configuration.java:178]: I am not seeing the need right now to make all the default values as separate constants. I think making this `""` is sufficient for now.

File: analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java (hunk 8 above)
Comment (dwalluck) [reply analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java:663]: I think the filename is a relative filename, relative to where we will place the file, so we have to resolve it first.

But you mean I am not calling the right API to retrieve the output directory to resolve against?

File: analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java (hunk 11 above)
Comment (rnc) [reply analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java:775]: Other files writing to the root directory need to be committed by repour - thats the difference.

File: common/src/main/java/org/jboss/gm/common/Configuration.java (hunk 10 above)
Comment (dwalluck) [reply common/src/main/java/org/jboss/gm/common/Configuration.java:178]: Also, with the keys, maybe it's not necessary if you have access to the configuration object, but I see tests setting properties and they need to know the name of the properties to set.

Comment (rnc) [reply common/src/main/java/org/jboss/gm/common/Configuration.java:178]: Right but we're not talking about keys here but the default value

File: analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java (hunk 8 above)
Comment (rnc) [reply analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java:663]: I'm confused?

File: analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java (hunk 12 above)
Comment (rnc) [reply analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java:751]: I think we're getting confused on terminology. I see in your report, a sub-module appearing as a 'dependency'?

File: analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java (hunk 8 above)
Comment (rnc) [reply analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java:663]: No I mean root directory but then inside `build` surely so repour doesn't commit it?

File: analyzer/src/functTest/java/org/jboss/gm/analyzer/alignment/SimpleProjectFunctionalTest.java (hunk 14 above)
Comment (rnc) [reply analyzer/src/functTest/java/org/jboss/gm/analyzer/alignment/SimpleProjectFunctionalTest.java:350]: Lets ignore this for now; it would be nice as an improvement but could be done in subsequent work (as PME doesn't sort either)

File: analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java (hunk 13 above)
Comment (rnc) [reply analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java:774]: Discussed on meeting ; it can be empty to disable the property

File: common/src/main/java/org/jboss/gm/common/Configuration.java (hunk 10 above)
Comment (dwalluck) [reply common/src/main/java/org/jboss/gm/common/Configuration.java:178]: OK, I left the key constant to refer to in the test, and removed the other, just using hardcoded `""`.

File: analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java (hunk 11 above)
Comment (dwalluck) [reply analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java:775]: Since the value is non-empty by default, `StringUtils.isEmpty(configuration.reportJsonOutputFile())` would be `true` iff the user has disabled it (via a system property). These is a test that checks that this is possible.

File: analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java (hunk 12 above)
Comment (dwalluck) [reply analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java:747]: OK, it's now printing `"------------------- project"` instead. To me, the text now looks good.

File: analyzer/src/functTest/java/org/jboss/gm/analyzer/alignment/SimpleProjectFunctionalTest.java [hunk 15]
```diff
@@ -211,4 +218,163 @@ public void verifySnapshotHandling() throws IOException, URISyntaxException, Man
...
+                        "    \"version\" : \"1.0.1\"%n" +
+                        "  },%n" +
+                        "  \"modules\" : [ {%n" +
+                        "    \"gav\" : {%n" +
+                        "      \"groupId\" : \"org.acme.gradle\",%n" +
+                        "      \"artifactId\" : \"root\",%n" +
+                        "      \"version\" : \"1.0.1.redhat-00002\",%n" +
+                        "      \"originalGAV\" : \"org.acme.gradle:root:1.0.1\"%n" +
+                        "    }%n" +
+                        "  }, {%n" +
```
dwalluck, line 252: The coordinates `org.acme.gradle:root` appear three times: 

1. Execution root.
2. Without modules (_root_ project).
3. With modules.

Is this correct?

In the text file, it appears twice (equivalent to 2 and 3 above).

analyzer/src/functTest/java/org/jboss/gm/analyzer/alignment/SimpleProjectFunctionalTest.java
```diff
@@ -41,9 +49,8 @@
 
 @RunWith(BMUnitRunner.class)
 public class SimpleProjectFunctionalTest extends AbstractWiremockTest {
-
     @Rule
-    public final SystemOutRule systemOutRule = new SystemOutRule().enableLog().muteForSuccessfulTests();
+    public final SystemOutRule systemOutRule = new SystemOutRule().enableLog()/*.muteForSuccessfulTests()*/;
```
dwalluck, line 53: TODO: revert

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
```diff
@@ -644,4 +651,116 @@ private void writeRepositorySettingsFile(Map<ArtifactRepository, Path> repositor
...
+                    }
+                });
+            }
+
+            if (!nonAligned.isEmpty()) {
+                for (ProjectVersionRef na : nonAligned) {
+                    logger.debug("Non-Aligned Dependencies : {}", na);
+                    append(builder, "\tNon-Aligned Dependencies : {}", na);
+                }
+            }
```
rnc, line 748: Question - is nonAligned reported on per sub-module/project basis or all together at the end? If the latter, could we add some spacing ?

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
rnc: Yeah looks good!

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
dwalluck: Good point---I think it needs to move up a line to 741 inside the `forEach` loop. It may be outside the loop now. Is there a multi-module project similar to simple-project that's ready to be used to test multi-module?

analyzer/src/functTest/java/org/jboss/gm/analyzer/alignment/SimpleProjectFunctionalTest.java
dwalluck: I think I may have asked this already on a resolved comment, but I changed some things to add the missing `originalGav` field and the `version` field was referencing the original instead of the new one. I am sure that you're verifying that the output of both files looks good as part of the review already.

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
dwalluck: Also, does it make sense to refactor this method to make it smaller?

rnc: Re "moving" - agreed.
Re "refactoring" - not sure how you can make it smaller without adding extra methods. For that matter, the block after ` if (cache.removeProject(project)) {` could be moved to its own method perhaps. Probably not essential to do at the moment IMHO.

Re "multi-module project" - many of the functional tests are multimodule - but you probably want one with some dependencies to try - you could look at the spring ones?

analyzer/src/functTest/java/org/jboss/gm/analyzer/alignment/SimpleProjectFunctionalTest.java
rnc: I've checked the textual output i.e.
```
[AlignmentTask.java:729] In module root with GAV io.undertow:undertow-core:2.0.15.Final found a replacement version of io.undertow:undertow-core:2.0.15.Final-redhat-00001
[AlignmentTask.java:754] 
------------------- project org.acme.gradle:root
	Project version : 1.0.1 --> 1.0.1.redhat-00002

------------------- project org.acme.gradle:root
	Project version : 1.0.1 --> 1.0.1.redhat-00002

	Dependencies : com.yammer.metrics:metrics-core:2.2.0 --> com.yammer.metrics:metrics-core:2.2.0-redhat-00001
	Dependencies : org.hibernate:hibernate-core:5.3.7.Final --> org.hibernate:hibernate-core:5.3.7.Final-redhat-00001
	Dependencies : io.undertow:undertow-core:2.0.15.Final --> io.undertow:undertow-core:2.0.15.Final-redhat-00001


[AlignmentTask.java:658] Writing to file /tmp/junit5061978604818961430/simple-project/build/alignmentReport.json
```
and that looks good - by the way shouldn't you also assert this format/contents?


As for the `alignmentReport.json`, you are correct this does look strange - I am seeing
```
{
    "executionRoot" : {
        "groupId" : "org.acme.gradle",
        "artifactId" : "root",
        "version" : "1.0.1.redhat-00002",
        "originalGAV" : "org.acme.gradle:root:1.0.1"
    },
    "modules" :
    [
        {
            "gav" :
            {
                "groupId" : "org.acme.gradle",
                "artifactId" : "root",
                "version" : "1.0.1.redhat-00002",
                "originalGAV" : "org.acme.gradle:root:1.0.1"
            }
        },
        {
            "gav" :
            {
                "groupId" : "org.acme.gradle",
                "artifactId" : "root",
                "version" : "1.0.1.redhat-00002",
                "originalGAV" : "org.acme.gradle:root:1.0.1"
            },
            "dependencies" :
            {
                "com.yammer.metrics:metrics-core:2.2.0" : {
                    "groupId" : "com.yammer.metrics",
                    "artifactId" : "metrics-core",
                    "version" : "2.2.0-redhat-00001"
                },
                "io.undertow:undertow-core:2.0.15.Final" : {
                    "groupId" : "io.undertow",
                    "artifactId" : "undertow-core",
                    "version" : "2.0.15.Final-redhat-00001"
                },
                "org.hibernate:hibernate-core:5.3.7.Final" : {
                    "groupId" : "org.hibernate",
                    "artifactId" : "hibernate-core",
                    "version" : "5.3.7.Final-redhat-00001"
                }
            }
        }
    ]
}
```
The first GAV looks wrong to me? Does PME do this?

rnc: I think the problem is in `if (!originalVersion.equals(newVersion)) {` block - you are adding an extra modulesitem in there. Further, I am not convinced on this version check - surely the execution root should **always** be added ?

dwalluck: > and that looks good - by the way shouldn't you also assert this format/contents?

It does, unless I misunderstand what you're asking. I could change it to a `String` instead of `List<String>`, but `String` wasn't saying which line had a problem. Can I upgrade `assertj-core` to latest version for better error messages when comparing strings?

>  Improvements
> 
> Change `isEqualTo` error message to follow pattern:
> 
> expected: "abc"
> but was : "bcd"

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
dwalluck: > Re "refactoring" - not sure how you can make it smaller without adding extra methods

I thought "refactoring" meant (possibly) adding/removing methods. Regardless, what I meant was moving some of the code to smaller methods instead of leaving a single large method.

analyzer/src/functTest/java/org/jboss/gm/analyzer/alignment/SimpleProjectFunctionalTest.java
rnc: I think we are talking at cross-purposes - I meant checking the console output _as well as_ the json string

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
dwalluck: > Question - is nonAligned reported on per sub-module/project basis or all together at the end? If the latter, could we add some spacing ?

OK, I fixed this to move the `nonAligned` inside the loop. It should be reported together with the sub-modules dependencies from what I see in PME. In PME, there is no line between the aligned and non-aligned dependencies, but there is a line after all of them are printed. Sometimes, there are two spacer lines, but it's hard to tell when this in intentional.

There is already a single spacer line after the dependencies. But, right now I think it might be possible that if nothing gets printed it still prints a spacer line.

rnc: Don't worry about the spacing - that was more if it _wasn't_ running inside the loop.

analyzer/src/functTest/java/org/jboss/gm/analyzer/alignment/SimpleProjectFunctionalTest.java
dwalluck: > I think we are talking at cross-purposes - I meant checking the console output _as well as_ the json string

I can check if `systemOutRule.getLog()` contains the same contents as the text file. In the first test, though, no text file is created, as that tests the default settings, but I can still check the log.

dwalluck: > Further, I am not convinced on this version check - surely the execution root should **always** be added ?

I think the `if` was only meant to be around the line that is printing it to the screen/text file. As far as the JSON, yes, it should always be there.

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
rnc: Ok the non-aligned dependencies from my manual test do now appear next to submodules.

I think (for future!) we have to do more work on isolating non-aligned down to direct dependencies etc.

analyzer/src/functTest/java/org/jboss/gm/analyzer/alignment/SimpleProjectFunctionalTest.java
rnc: I see the new tests - great! 

I still don't think the block `if (configuration.versionModificationEnabled()) {` is correct - as if versionModification for the project is disabled (for some reason) the entire report is lost. And I think you missed my comment about the incorrect json, and potentially why it was incorrect?

analyzer/src/functTest/java/org/jboss/gm/analyzer/alignment/SimpleProjectFunctionalTest.java
rnc: Shall we change this now?

analyzer/src/functTest/java/org/jboss/gm/analyzer/alignment/SimpleProjectFunctionalTest.java
dwalluck: > The first GAV looks wrong to me? Does PME do this? In fact, could that JSON even be read back in - as it doesn't look like it conforms?

I thought it was strange that `org.acme.gradle:root:1.0.1-redhat-00002` is shown twice (as `module` in JSON report and `project` in text file).

The first one is coming from `cache.getModel()` which I think is `project.getRootProject()` and the second is coming from processing the dependencies `cache.getDependencies()`. You can see it appears twice in the text file as well.

dwalluck: > if (configuration.versionModificationEnabled()) {

Actually, there is no tests for this option currently. Getting at the configuration requires bmunit? See <https://github.com/project-ncl/gradle-manipulator/blob/36a6bef11661cd83042bb1cc63a86a38dd1b1c13/analyzer/src/functTest/java/org/jboss/gm/analyzer/alignment/SimpleExistingManipulationProjectFunctionalTest.java#L90-L105>.

I think it is my mistake in thinking that when this option is enabled no version change is expected to occur, but there are still potential overrides, etc. So, this just refers to the intra-project versions?

analyzer/src/functTest/java/org/jboss/gm/analyzer/alignment/SimpleProjectFunctionalTest.java
dwalluck: I planned to once all outstanding issues were fixed. This helps to see issues related to the reports right on the console.

analyzer/src/functTest/java/org/jboss/gm/analyzer/alignment/SimpleProjectFunctionalTest.java
rnc: | So, this just refers to the intra-project versions?
Yes.

| I thought it was strange that org.acme.gradle:root:1.0.1-redhat-00002 is shown twice
Its appearing 3 times

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
```diff
@@ -644,4 +650,130 @@ private void writeRepositorySettingsFile(Map<ArtifactRepository, Path> repositor
...
+        append(builder, "------------------- project {}", originalGa);
+        final String originalVersion = alignmentModel.getOriginalVersion();
+        final String newVersion = alignmentModel.getVersion();
+        final PME jsonReport = new PME();
+        final List<ModulesItem> modules = jsonReport.getModules();
+        final ModulesItem rootModule = new ModulesItem();
+        final String gav = originalGa + ":" + newVersion;
+        final ProjectVersionRef pvr = SimpleProjectVersionRef.parse(gav);
+        rootModule.getGav().setPVR(pvr);
+        modules.add(rootModule);
```
dwalluck, line 679: Is this necessary? It must also set this same GAV later around line 720.

analyzer/src/functTest/java/org/jboss/gm/analyzer/alignment/SimpleProjectFunctionalTest.java
dwalluck: If you count `executionRoot`, but I think this one is correct. It appears only twice in the text file. I made another comment on the code itself which might be the cause.

analyzer/src/functTest/java/org/jboss/gm/analyzer/alignment/SimpleProjectFunctionalTest.java
rnc: Ok

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
rnc: In my previous analysis, before you force pushed,  I had diagnosed which was the problem. I'm on phone now so am not sure

analyzer/src/functTest/java/org/jboss/gm/analyzer/alignment/SimpleProjectFunctionalTest.java
dwalluck: > I still don't think the block `if (configuration.versionModificationEnabled()) {` is correct - as if versionModification for the project is disabled (for some reason) the entire report is lost. And I think you missed my comment about the incorrect json, and potentially why it was incorrect?

I have removed the `if (configuration.versionModificationEnabled())` entirely from that section meaning that the `executionRoot` is always set.

dwalluck: > I think the problem is in `if (!originalVersion.equals(newVersion)) {` block - you are adding an extra modulesitem in there. Further, I am not convinced on this version check - surely the execution root should **always** be added ?

I have changed this so that now the `originalGav` field is set only if `!originalVersion.equals(newVersion)`, but everything else is created/set unconditionally.

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
dwalluck: > In my previous analysis, before you force pushed, I had diagnosed which was the problem. I'm on phone now so am not sure

You meant this one?

> I think the problem is in if (!originalVersion.equals(newVersion)) { block - you are adding an extra modulesitem in there. Further, I am not convinced on this version check - surely the execution root should always be added ?

I'll look at the code again. I may have missed the fact that these were two different sentences talking about two different things.

Also in response to:

> The first GAV looks wrong to me? Does PME do this? In fact, could that JSON even be read back in - as it doesn't look like it conforms?

The JSON should get read in during the tests, so it does parse, at least. I think it's the test that's wrong right now, and should fail, then, once this is fixed and itself need to be fixed.

rnc (approved):
I've approved, but could you change that `TODO` as a final change please.

PR #280, opened 2021-04-22
NCL-6520 Expose Translator REST API for Groovy scripting

rnc:


cli/src/main/java/org/jboss/gm/cli/Main.java
```diff
@@ -298,5 +298,17 @@ public int handleExecutionException(Exception ex, CommandLine commandLine, Comma
             this.exception = ex;
             return 1;
         }
+
+        public Exception getException() {
+            if (exception != null) {
+                if ("Caught exception running build".equals(exception.getMessage())) {
+                    exception = (Exception) exception.getCause();
+                }
+                if ("org.gradle.internal.exceptions.LocationAwareException".equals(exception.getClass().getName())) {
```
dwalluck, line 307: I think it's more correct and more performant to use `if (exception instanceof org.gradle.internal.exceptions.LocationAwareException)` here, although I find this pattern in two other places in the code as well.

analyzer/src/functTest/java/org/jboss/gm/analyzer/alignment/SimpleProjectWithCustomGroovyScriptFunctionalTest.java
```diff
@@ -93,13 +93,19 @@ public void verifyGroovyInjection() throws IOException, URISyntaxException, Mani
...
         assertTrue(systemOutRule.getLog().contains("original version is 1.0.1" + System.lineSeparator()));
-
+        assertThat(FileUtils.readFileToString(new File(projectRoot, "build.gradle"), Charset.defaultCharset()))
+                .satisfies(s -> {
+                    assertTrue(s.contains("classpath \"org.hibernate:hibernate-core:5.3.7.Final-redhat-00001"));
+                    assertTrue(s.contains("publishing {\n" +
+                            "  publications {\n" +
+                            "    nebula {\n" +
+                            "      artifact(\"build/distributions/${project.name}-${version}.zip\")\n"));
+                });
```
dwalluck, line 103: This and the next, it seems you could simplify the logic (and get better error messages) with

```
assertThat(s).contains("foo").contains("bar").doesNotContain("baz");
```

cli/src/main/java/org/jboss/gm/cli/Main.java
rnc: Except that class isn't in the module classpath

analyzer/src/functTest/java/org/jboss/gm/analyzer/alignment/SimpleProjectWithCustomGroovyScriptFunctionalTest.java
rnc: I'd be happy with the damn tests passing!

cli/src/main/java/org/jboss/gm/cli/Main.java
rnc: Also it's on a fail path in the code - the app is about to exit so it really shouldn't matter at that point.

rnc:
@dwalluck While the different attempts at caching (I've kept the commits separate for now so you can look at them) didn't solve the problem, the final [`00de400` (#280)](https://github.com/project-ncl/gradle-manipulator/pull/280/commits/00de4000983b21c08891acbe9e7bd3942533b237) did.

dwalluck:
> @dwalluck While the different attempts at caching (I've kept the commits separate for now so you can look at them) didn't solve the problem, the final [`00de400` (#280)](https://github.com/project-ncl/gradle-manipulator/pull/280/commits/00de4000983b21c08891acbe9e7bd3942533b237) did.
> 
> Unless you have any comments, I can squash all this and merge it.

Obviously, you note it is a hack. As for the real cause, as `Unirest.config()` is `static`, I believe the actual cause may be https://github.com/project-ncl/gradle-manipulator/blob/e96a7a1e60c621e4cfb8674718b3fda512ec4c2d/cli/src/test/java/org/jboss/gm/cli/DifferentJVMTest.java#L68

and `DifferentJVMTest` does run before before `MainTest`.

I would like to see if this change is even needed if `DifferentJVMTest` is fixed. In particular, I think there should be an `@AfterClass` method that calls `Unirest.config().reset()` because the test has messed with some static state and not reset it back to the original state for the other tests (very similar to how a JUnit `@Rule` would work to prevent similar issues).

rnc:
Oh fantastic spot! I missed that!

PR #283, opened 2021-04-22
NCL-6471: Always overwrite existing marker files during alignment

dwalluck:


common/src/main/java/org/jboss/gm/common/utils/FileUtils.java
```diff
@@ -36,19 +36,20 @@ public static String getLastLine(File target) throws IOException {
...
-    public static String getFirstLine(List<String> lines) throws ManipulationException {
+    public static String getFirstLine(List<String> lines) {
         for (String line : lines) {
             if (StringUtils.isNotBlank(line)) {
                 return line;
             }
         }
-        throw new ManipulationException("Unable to find a non blank line in the collection");
+
+        return StringUtils.EMPTY;
```
dwalluck, line 52: This change is to allow calling this method on an empty file (which my test creates). I don't quite understand why it just doesn't return

```
lines.isEmpty() ? StringUtils.EMPTY : lines.get(0);
```

but that breaks several of the current tests if done that way.

rnc: You can't return `lines.get(0)` as you don't know if that is blank or not - you need to iterate through to find the first non-blank one.

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
```diff
@@ -383,8 +376,9 @@ private void updateAllExtraGradleFilesWithGmeRepos() throws IOException, Manipul
                 DirectoryFileFilter.DIRECTORY);
         for (File extraGradleScript : extraGradleScripts) {
             final List<String> lines = FileUtils.readLines(extraGradleScript, Charset.defaultCharset());
+            final String first = org.jboss.gm.common.utils.FileUtils.getFirstLine(lines);
 
-            if (!APPLY_GME_REPOS.equals(org.jboss.gm.common.utils.FileUtils.getFirstLine(lines))) {
+            if (!APPLY_GME_REPOS.equals(first)) {
```
rnc, line 381: Why this change?

common/src/main/java/org/jboss/gm/common/utils/FileUtils.java
rnc: I'm not convinced by this change ; the file its failing on is `gradle/gme-repos.gradle` which is one that should be overwritten. I don't think we should be changing utility functions to make tests work.

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
```diff
@@ -81,16 +81,16 @@
...
+    public static final String INJECT_GME_START_KOTLIN = "buildscript { project.apply { from(\"" + GME + "\") } }";
+    public static final String GME_PLUGINCONFIGS = "gme-pluginconfigs.gradle";
+    public static final String INJECT_GME_END = "apply from: \"" + GME_PLUGINCONFIGS + "\"";
+    public static final String INJECT_GME_END_KOTLIN = "project.apply { from(\"" + GME_PLUGINCONFIGS + "\") }";
     public static final String GRADLE = "gradle";
     public static final String GME_REPOS = "gme-repos.gradle";
-    public static final String APPLY_GME_REPOS = "buildscript { apply from: new File(buildscript.getSourceFile().getParentFile(),\"gme-repos.gradle\"), to: buildscript }";
-    public static final String GME_PLUGINCONFIGS = "gme-pluginconfigs.gradle";
+    public static final String APPLY_GME_REPOS = "buildscript { apply from: new File(buildscript.getSourceFile().getParentFile(), "
+            + GME_REPOS + "\"), to: buildscript }";
```
rnc, line 93: This has broken the injection. Please revert these changes

rnc (changes requested):
This isn't right - we now end up with 
```
  +buildscript { apply from: new File(buildscript.getSourceFile().getParentFile(), gme-repos.gradle"), to: buildscript }
  +
  +
  +buildscript { apply from: new File(buildscript.getSourceFile().getParentFile(), "gme-repos.gradle"), to: buildscript }
```
in modified files.

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
dwalluck: There are constants defined right there that aren't used. If the constants are ever changed, the script texts will refer to the wrong file. I just tried to replace the text there with the constant representing the exact same text, unless I made a typo somewhere.

Also, if something is failing, how can all the tests pass?

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
dwalluck: I will revert it. Originally, I needed to pull it out to a variable for some other reason that's no longer in the code.

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
dwalluck: I see, there's a missing `"`. But there must be a test that injects and then checks that the injection is there? If not, I could add one.

common/src/main/java/org/jboss/gm/common/utils/FileUtils.java
dwalluck: Specifically, the change is supposed to allow injecting into an empty file without failing. I did not understand why blank lines were relevant in the first place and why it is not just placed at the very top, blank lines or not.

In any case, I can revert this if I start in my test off with a non-blank script. What is the most minimal script it is looking for, then? As far as the test goes, it was easiest to start with an empty file and then check if the file is not empty as part of the test.

But I wasn't sure it was really a "feature" to be failing in the case of an empty file or file with just whitespace regardless of the test.

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
dwalluck: Reverted.

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
```diff
@@ -379,7 +378,8 @@ private void updateAllExtraGradleFilesWithGmeRepos() throws IOException, Manipul
         if (!gradleScriptsDirectory.exists()) {
             return;
         }
-        final Collection<File> extraGradleScripts = FileUtils.listFiles(gradleScriptsDirectory, new SuffixFileFilter(".gradle"),
+        final Collection<File> extraGradleScripts = FileUtils.listFiles(gradleScriptsDirectory,
+                FileFilterUtils.and(new SuffixFileFilter(".gradle"), new NotFileFilter(new NameFileFilter(GME_REPOS))),
```
dwalluck, line 382: Proposed fixed is to exclude any existing `GME_REPOS` file from update as it will be overwritten later anyway.

common/src/main/java/org/jboss/gm/common/utils/FileUtils.java
dwalluck: Reverted.

rnc (approved):

PR #284, opened 2021-04-26
[NCL-6167] Use new PME/DA endpoint parameters

thescouser89:
The new PME (4.3) and DA (2.x) introduces 2 new parameters:

- brewPullActive (default: false)
- mode (default: "")

The 'brewPullActive' flag is set on whether to also consider artifacts
from Brew during alignment.

The mode is used to specify to which aligned versions to align to. Currently
there are 4 modes:

- PERSISTENT (permanent aligned builds)
- TEMPORARY (temporary aligned builds)
- SERVICE (service aligned builds)
- SERVICE-TEMPORARY (temporary service aligned builds)

Those modes and the corresponding aligned version suffixes are
configured on the Dependency Analysis service.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -129,7 +129,7 @@ subprojects {
     extra["logbackVersion"] = "1.2.3"
     extra["mavenVersion"] = "3.5.0"
     extra["ownerVersion"] = "1.0.12"
-    extra["pmeVersion"] = "4.2"
+    extra["pmeVersion"] = "4.3"
     extra["slf4jVersion"] = "1.7.30"
     extra["systemRulesVersion"] = "1.19.0"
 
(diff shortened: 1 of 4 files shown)


rnc:
@thescouser89 Thanks! I know this is draft :-) but once done we'll need an update to the gh-pages branch as well I think

build.gradle.kts
```diff
@@ -129,7 +129,7 @@ subprojects {
     extra["logbackVersion"] = "1.2.3"
     extra["mavenVersion"] = "3.5.0"
     extra["ownerVersion"] = "1.0.12"
-    extra["pmeVersion"] = "4.2"
+    extra["pmeVersion"] = "4.3-SNAPSHOT"
```
rnc, line 132: This can be updated to 4.3 now please

rnc:
@thescouser89 As part of the move to 4.3 could you remove the `TODO` from https://github.com/project-ncl/gradle-manipulator/blob/master/common/src/main/java/org/jboss/gm/common/groovy/BaseScript.java#L107 please?

thescouser89:
Aye codecov isn't happy. Let me know if you want me to add more tests or something.

rnc (approved):

PR #289, opened 2021-04-30
Create dependabot.yml

jwillemsen:


rnc:
Hi @jwillemsen ! Thanks for this PR. There are pros/cons for continual dependency updates. Note that this project is already configured for notifications for CVE updates. Also I think the package ecosystem is wrong. Can you expand upon why you submitted this please? Are you currently using GME?

jwillemsen:
I am not using this but did notice you are using github actions, this file helps to keep your action up to date, something we use frequently

jwillemsen:
It keeps .github/workflows/gradle.yml up to date

rnc:
> It keeps .github/workflows/gradle.yml up to date

Yes, I understand that ; it would be only for major version changes I guess given how we have coded our action usage?

jwillemsen:
> Yes, I understand that ; it would be only for major version changes I guess given how we have coded our action usage?

Yes, as far as I can tell it is only for a major version upgrade

dwalluck:
You can see from the file that I created https://raw.githubusercontent.com/project-ncl/build-finder/master/.github/dependabot.yml that I have two package ecosystems, "github-actions" and "maven". But, perhaps unfortunately, it will update, say, v2 to v2.1.4 as in this example project-ncl/build-finder#361, which means that after the change you would have to rely on dependabot for updates and can no longer use the major version "feature".

rnc:
> You can see from the file that I created [project-ncl/build-finder@`master`/.github/dependabot.yml (raw)](https://raw.githubusercontent.com/project-ncl/build-finder/master/.github/dependabot.yml) that I have two package ecosystems, "github-actions" and "maven". But, perhaps unfortunately, it will update, say, v2 to v2.1.4 as in this example [project-ncl/build-finder#361](https://github.com/project-ncl/build-finder/pull/361), which means that after the change you would have to rely on dependabot for updates and can no longer use the major version "feature".

Thanks for the input @dwalluck. That does not sound good at all, as I would prefer to use latter syntax.

rnc:
@jwillemsen After @dwalluck demonstrated the issue, I am afraid I am going to reject this. Currently with the Action usage of `@v1` we will automatically get all 1.x versions. For a major version change, which might require code changes anyway, admittedly that has to be done manually.

PR #290, opened 2021-04-30
NCL-6543: Remove "Testing on a real project" section

dwalluck:
Part 1. Part 2 moves this section to the `gh-pages` branch, assuming file `index.md`?


rnc:
Looks good .... I wonder if we should say instead of `Documentation for the project can be found` but `Documentation and usage instructions for the project can be found` ?

dwalluck:
> Looks good .... I wonder if we should say instead of `Documentation for the project can be found` but `Documentation and usage instructions for the project may be found` ?

Where?

rnc:
> > Looks good .... I wonder if we should say instead of `Documentation for the project can be found` but `Documentation and usage instructions for the project may be found` ?
> 
> Where?

In the readme.md

rnc (approved):

PR #292, opened 2021-05-07
Support NO_COLOR. Remove MDC support (NCL-5068)

rnc:


rnc:
@thescouser89 @dwalluck Adding `NO_COLOR` support to GME so that if repour sets that environment variable its automatically picked up.

cli/src/main/java/org/jboss/gm/cli/Main.java
```diff
@@ -197,7 +198,7 @@ private void executeGradle() throws ManipulationException {
                     .map(entry -> "-D" + entry.getKey() + '=' + entry.getValue())
                     .collect(Collectors.toSet());
 
-            if (colour) {
+            if (colour && StringUtils.isEmpty(System.getenv("NO_COLOR"))) {
```
thescouser89, line 201: There's a typo in the 'color' name :stuck_out_tongue:

dwalluck: Just for fun, it seems we have too: `NO_COLOR`, `build.setColorOutput`, `-DloggingColours=false`. You're not going to be able to get it 100% consistent, anyway.

dwalluck: I thought that both British and Canadian spellings agreed on the "our" ending and I haven't so far suggested that native speakers change the spelling to American, but I wonder if our tools shouldn't support aliases for every occurrence (but I personally haven't felt like making such a patch).

thescouser89 (approved):
LGTM!

PR #294, opened 2021-05-17
NCL-6109: Add jars: common, common-javadoc, common-sources, common-te…

dwalluck:
…st-fixtures

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -246,61 +246,10 @@ subprojects {
                         }
                     }
 
-                    pom {
-                        name.set("Gradle Manipulation Extension")
-                        description.set("A tool to work with ProjectNCL to manipulate Gradle builds.")
-                        url.set("https://github.com/project-ncl/gradle-manipulator")
-                        packaging = "jar"
-
-                        licenses {
-                            license {
-                                name.set("The Apache License, Version 2.0")
-                                url.set("http://www.apache.org/licenses/LICENSE-2.0.txt")
-                            }
-                        }
-                        developers {
-                            developer {
-                                id.set("geoand")
-                                name.set("Georgios Andrianakis")
-                                email.set("gandrian@redhat.com")
-                            }
-                            developer {
-                                id.set("rnc")
-                                name.set("Nick Cross")
-                                email.set("ncross@redhat.com")
-                            }
-                            developer {
-                                id.set("TomasHofman")
-                                name.set("Tomas Hofman")
-                                email.set("thofman@redhat.com")
-                            }
-                            developer {
-                                id.set("metacosm")
-                                name.set("Chris Laprun")
-                                email.set("claprun@redhat.com")
-                            }
-                        }
-                        scm {
-                            connection.set("scm:git:http://github.com/project-ncl/gradle-manipulator.git")
-                            developerConnection.set("scm:git:git@github.com:project-ncl/gradle-manipulator.git")
-                            url.set("https://github.com/project-ncl/gradle-manipulator")
-                        }
-                    }
+                    generatePom()
                 }
             }
-            repositories {
-                if (isReleaseBuild) {
-                    maven {
-                        name = "sonatype-nexus-staging"
-                        url = uri("https://oss.sonatype.org/service/local/staging/deploy/maven2")
-                    }
-                } else {
-                    maven {
-                        name = "sonatype-nexus-snapshots"
-                        url = uri("https://oss.sonatype.org/content/repositories/snapshots")
-                    }
-                }
-            }
+            generateRepositories(this, this@Build_gradle)
         }
 
         if (isReleaseBuild) {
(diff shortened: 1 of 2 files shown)


build.gradle.kts
```diff
@@ -143,11 +143,16 @@ subprojects {
         enabled = false
     }
 
-    if (project.name == "common" || project.name == "cli") {
+
+    // Don't apply Gradle plugin code to the cli tool
+    if (project.name == "cli") {
         apply(plugin = "java-library")
```
rnc, line 149: Why aren't you applying the java-library to common?

build.gradle.kts
```diff
@@ -156,147 +161,146 @@ subprojects {
...
+     * Another great source of information is the configuration of the shadow plugin itself:
+     * https://github.com/johnrengelman/shadow/blob/master/build.gradle
+     */
+
+    // make build task depend on shadowJar
+    val build: DefaultTask by tasks
+    val shadowJar = tasks["shadowJar"] as ShadowJar
+    build.dependsOn(shadowJar)
+
+    tasks.withType<ShadowJar> {
```
rnc, line 183: Doesn't this if structure mean that the common library will be published as a Shadow publication?

build.gradle.kts
```diff
@@ -156,147 +161,146 @@ subprojects {
...
-                    if (project.name == "analyzer") {
-                        artifact("${sourceSets.main.get().output.resourcesDir}/analyzer-init.gradle") {
-                            classifier = "init"
-                            extension = "gradle"
-                        }
+
+    // configure publishing of the shadowJar
+    configure<PublishingExtension> {
+        publications {
+            create<MavenPublication>("shadow") {
```
rnc, line 227: As per above re comment for shadow - I think a different publication will be needed for common versus cli/analyzer/etc?

build.gradle.kts
```diff
@@ -156,147 +161,146 @@ subprojects {
...
-                    attributes["Build-Timestamp"] = SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSZ").format(Date())
-                    attributes["Scm-Revision"] = Grgit.open(mapOf("currentDir" to project.rootDir)).use { g -> g.head().id }
-                    attributes["Created-By"] = "Gradle ${gradle.gradleVersion}"
-                    attributes["Build-Jdk"] = System.getProperty("java.version") + " ; " + System.getProperty("java.vendor") + " ; " + System.getProperty("java.vm.version")
-                    attributes["Build-OS"] = System.getProperty("os.name") + " ; " + System.getProperty("os.arch") + " ; " + System.getProperty("os.version")
-                    attributes["Implementation-Version"] = "${project.version}"
-                }
+    apply(plugin = "signing")
+    apply(plugin = "maven-publish")
+    apply(plugin = "com.github.johnrengelman.shadow")
```
rnc, line 167: Why does common need a shadow plugin?

rnc (changes requested):
See inline comments - I am not convinced this looks correct (and the CI is failing)

build.gradle.kts
```diff
@@ -295,8 +296,91 @@ subprojects {
...
+
+        val sourcesJar by tasks.registering(Jar::class) {
+            archiveClassifier.set("sources")
+            from(sourceSets.main.get().allSource)
+        }
+
+        val javadocJar by tasks.registering(Jar::class) {
+            archiveClassifier.set("javadoc")
+            from(tasks["javadoc"])
+        }
```
rnc, line 312: Given the above block (300->312) is almost a duplicate of the lines around 162 I would think we could improve the if blocks?

build.gradle.kts
```diff
@@ -295,8 +296,91 @@ subprojects {
...
+                            }
+                        }
+                        scm {
+                            connection.set("scm:git:http://github.com/project-ncl/gradle-manipulator.git")
+                            developerConnection.set("scm:git:git@github.com:project-ncl/gradle-manipulator.git")
+                            url.set("https://github.com/project-ncl/gradle-manipulator")
+                        }
+                    }
+                }
+            }
```
rnc, line 361: I am not seeing a common jar:
```
NCL-6109(dwalluck/NCL-6109)|…6⚑4] 🎩❯ l ~/.m2/repository/org/jboss/gm/common/2.8-SNAPSHOT                                                                               14:50
total 524
-rw-rw-r--. 1 rnc rnc 497518 May 24 14:45 common-2.8-SNAPSHOT-javadoc.jar
-rw-rw-r--. 1 rnc rnc   1592 May 24 14:45 common-2.8-SNAPSHOT.pom
-rw-rw-r--. 1 rnc rnc  25646 May 24 14:45 common-2.8-SNAPSHOT-sources.jar
-rw-rw-r--. 1 rnc rnc    942 May 24 14:45 maven-metadata-local.xml
```

While the local jar does look *much* better, I do wonder whether we can avoid the large amount of duplication here?

rnc: Ok now I am seeing a jar.

build.gradle.kts
rnc: I am still seeing what looks like duplicate definitions of ```
        apply(plugin = "signing")
        apply(plugin = "maven-publish")
        apply(plugin = "net.linguica.maven-settings")
``` and sourcesJar, javadocJar, etc ? 


So while functionally it looks correct, I think it could be optimised?

rnc (approved):

PR #295, opened 2021-05-21
NCL-6109: Fix javadoc warnings and errors in common module

dwalluck:


dwalluck:
This isn't building the common jar yet, but I though that the javadoc changes ought to be separate from the build changes,

PR #302, opened 2021-06-22
Fix some resource leaks

dwalluck:


manipulation/src/functTest/resources/elasticsearch/buildSrc/src/main/minimumRuntime/org/elasticsearch/gradle/VersionProperties.java
```diff
@@ -35,17 +35,19 @@ public static String getLucene() {
...
+                throw new IllegalStateException("/version.properties resource missing");
+            }
+
+            versionProps.load(propsStream);
         } catch (IOException e) {
             throw new IllegalStateException("Failed to load version properties", e);
         }
-        return props;
+
+        return versionProps;
```
rnc, line 50: You could make a PR to elasticsearch if that code is still the same - that is where that test case was sourced from!

rnc (approved):

manipulation/src/functTest/resources/elasticsearch/buildSrc/src/main/minimumRuntime/org/elasticsearch/gradle/VersionProperties.java
dwalluck: I will look. I also changed the name of the `props` variable as you saw. I think it was shadowing a field of the same name which was a little confusing.

PR #303, opened 2021-06-22
Escape backslashes on Windows in repositories/flatDir/dirs

dwalluck:


dwalluck:
This is just a quick fix to make `MainTest` work with Windows paths.

Oddly, for `dirs`, using `/` doesn't seem to work either as Gradle thinks the path is relative, so you have to use a double backslash`\\`.

Also, it fixes another potential issue by using `toUri()` instead of `getPath()`.

rnc:
Are we running this at all on Windows? I just wonder why we're going to the effort?

PR #304, opened 2021-06-23
NCL-6645: Revert method reference in lambda

dwalluck:


dwalluck:
There was an erroneous IDEA quick fix applied in my NCL-6601 pull request.

dwalluck:
It was not caught because we don't have ITs for GME 4.10 yet.

rnc (approved):

PR #305, opened 2021-06-27
NCL-6647 Logging improvements. CLI debug improvement

rnc:


cli/src/main/java/org/jboss/gm/cli/Main.java
```diff
@@ -198,6 +199,8 @@ private void executeGradle() throws ManipulationException {
             Set<String> jvmArgs = jvmPropertyParams.entrySet().stream()
                     .map(entry -> "-D" + entry.getKey() + '=' + entry.getValue())
                     .collect(Collectors.toCollection(LinkedHashSet::new));
+            jvmArgs.addAll(ManagementFactory.getRuntimeMXBean().getInputArguments().stream()
+                    .filter(s -> s.startsWith("-Xdebug") || s.startsWith("-Xrunjdwp")).collect(Collectors.toSet()));
```
dwalluck, line 203: Can you explain the reason for this line?

common/src/main/java/org/jboss/gm/common/logging/FilteringCustomLogger.java
```diff
@@ -42,13 +51,12 @@ private FilteringCustomLogger(OutputEventListener outputEventListener) {
...
-                // Can't use logger to output a warning as causes stack overflow.
-                // TODO: Remove?
-                // System.err.println("Unknown event using category " + logEvent.getCategory() + " : ");
+            if (logEvent.getCategory().startsWith("org.commonjava.maven.ext")) {
+                delegate.onOutput(new LogEvent(logEvent.getTimestamp(), logEvent.getCategory(), LogLevel.LIFECYCLE,
+                        logEvent.getMessage(), logEvent.getThrowable()));
+            } else {
                 delegate.onOutput(event);
             }
-            delegate.onOutput(event);
```
dwalluck, line 51: It looks like there was fall-through here which would make it print twice (on line 49 and again on line 51 if the expression evaluates to `true`) and was meant to have an `else` there.

cli/src/main/java/org/jboss/gm/cli/Main.java
rnc: If I'm trying to debug a GME run, if I do it through the CLI, it doesn't debug the spawned Gradle process - which is what I really wanted in the first place. So this propagates those CLI parameters so that the Gradle alignment run may be debugged.

common/src/main/java/org/jboss/gm/common/logging/FilteringCustomLogger.java
rnc: Yes, historically it had a different code path within the block, but was changed and wasn't updated.

PR #310, opened 2021-07-06
NCL-6050: Remove usage of SystemProperties for Gradle 4.10 compatibility

dwalluck:
The method `isStandardProperty()` does not exist in Gradle 4.10. The closest old method is `getStandardProperties()`, but this was removed in Gradle 5.0.

Regardless, because we already check for the property prefixes `java.`, `sun.`, and `user.`, the majority of properties handled by that method are duplicates anyway, so we can simplify this code.

### DIFF
--- a/manipulation/src/functTest/java/org/jboss/gm/manipulation/TestUtils.java
+++ b/manipulation/src/functTest/java/org/jboss/gm/manipulation/TestUtils.java
@@ -18,15 +18,21 @@ import org.apache.maven.model.io.xpp3.MavenXpp3Reader;
 import org.assertj.core.groups.Tuple;
 import org.codehaus.plexus.util.xml.pull.XmlPullParserException;
 import org.gradle.internal.Pair;
-import org.gradle.internal.SystemProperties;
 import org.gradle.testkit.runner.GradleRunner;
 import org.gradle.testkit.runner.internal.DefaultGradleRunner;
 import org.jboss.gm.common.model.ManipulationModel;
 
+import static org.apache.commons.lang3.StringUtils.equalsAny;
+import static org.apache.commons.lang3.StringUtils.startsWithAny;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.tuple;
 
 public final class TestUtils {
+
+    private static final String[] PROPERTY_PREFIXES = { "java.", "user.", "sun." };
+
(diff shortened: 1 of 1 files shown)


manipulation/build.gradle.kts
```diff
@@ -43,6 +43,7 @@ dependencies {
 
     testRuntimeOnly("commons-io:commons-io:${project.extra.get("commonsVersion")}")
     testImplementation("junit:junit:${project.extra.get("junitVersion")}")
+    testImplementation("org.apache.commons:commons-lang3:${project.extra.get("commonsLang3Version")}")
```
dwalluck, line 46: This version (3.5) is already on the classpath due to some other dependency, but not explicitly required. It is needed because 2.x which we do already include does not contain the `equalsAny` method.

build.gradle.kts
```diff
@@ -128,6 +128,7 @@ subprojects {
     extra["bytemanVersion"] = "4.0.15"
     extra["commonsBeanVersion"] = "1.9.4"
     extra["commonsVersion"] = "2.6"
+    extra["commonsLang3Version"] = "3.5"
```
dwalluck, line 131: This is an older version, but see note below as to why this version is used.

rnc: I see both 3.5 and 3.7 on the test classpaths (compile/runtime)

manipulation/build.gradle.kts
rnc: If this is *already* on the (test runtime) classpath why do we need to explicitly state it?

build.gradle.kts
dwalluck: I was stating the version because as a rule I suppose we are not supposed to rely on transitive dependencies in case they change out from under us.

I forgot how to get the exact dependency tree with gradle, but when I stepped into the class with IDEA, it showed the 3.5 jar. However, if we were to state it anyway, why not use the current latest version? It's just a test dependency anyway so it won't affect users.

rnc: TBH we are probably using lots of transitive without realising! We don't have a similar plugin to what we have in maven to enforce explicitly mentioning it. Honestly, I wouldn't bother adding those lines especially as its just for tests.

dwalluck: OK, I will remove the build file changes in that case and just use the classes (already) on the Gradle class path.

rnc (approved):

PR #311, opened 2021-07-06
NCL-6050: Fix reference to DefaultProjectDependencyConstraint when co…

dwalluck:
…mpiling with Gradle 4.10

While `d instanceof DefaultProjectDependencyConstraint` works at runtime under Gradle 4.10, it does not work at compile time as the class does not exist.

This effectively changes  `d instanceof DefaultProjectDependencyConstraint` to `Class.forName("DefaultProjectDependencyConstraint").isInstance(d)` instead.

### DIFF
--- a/analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
+++ b/analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
@@ -51,7 +51,6 @@ import org.gradle.api.artifacts.ResolvedDependency;
 import org.gradle.api.artifacts.UnresolvedDependency;
 import org.gradle.api.artifacts.repositories.ArtifactRepository;
 import org.gradle.api.internal.artifacts.configurations.ConflictResolution;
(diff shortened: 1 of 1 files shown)


rnc:
Looks fine but can we add a comment to the code why its using reflection please?

dwalluck:
> Looks fine but can we add a comment to the code why its using reflection please?

Sure, if the explanation here on the PR is OK, should I put it right near where I have the `static` block, add some javadoc to the `isDefaultProjectDependencyConstraint` method, or both?

By the way, which test is already testing this, or should I have another unit test somewhere?

rnc:
> > Looks fine but can we add a comment to the code why its using reflection please?
> 
> Sure, if the explanation here on the PR is OK, should I put it right near where I have the `static` block, add some javadoc to the `isDefaultProjectDependencyConstraint` method, or both?

Whichever you prefer ; as long as there is something for posterity explaining the reasoning.

> By the way, which test is already testing this, or should I have another unit test somewhere?

Not sure, would have to track back to look when it was introduced.

rnc (approved):

PR #312, opened 2021-07-06
NCL-6050: Unify Main exception handling due to Gradle 4.10

dwalluck:
Currently we catch `BuildException` and `GradleConnectionException` separately.

Both exceptions are instances of `RuntimeException` and we don't really need to treat them separately, so this unifies both to use the same code. Under Gradle 4.10, the code path is different. Code under test which expects  `GradleConnectionException` under 5.x has a different exception being thrown under 4.10 which makes the test too specific.

However, we had some reason for doing this, was it to use `e.getCause()`?

Also, pass the exception being thrown back into the thrown `ManipulationException` so that we don't lose that information.

### DIFF
--- a/cli/src/main/java/org/jboss/gm/cli/Main.java
+++ b/cli/src/main/java/org/jboss/gm/cli/Main.java
@@ -20,7 +20,6 @@ import org.apache.commons.lang.StringUtils;
 import org.commonjava.maven.ext.common.ManipulationException;
 import org.commonjava.maven.ext.core.groovy.InvocationStage;
 import org.gradle.internal.Pair;
-import org.gradle.tooling.BuildException;
 import org.gradle.tooling.BuildLauncher;
 import org.gradle.tooling.GradleConnectionException;
 import org.gradle.tooling.GradleConnector;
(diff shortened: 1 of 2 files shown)


rnc:
This was changed in https://github.com/project-ncl/gradle-manipulator/pull/246

This does change the logging and how exceptions are propagated. I think I'd like to understand the difference with 4.10 a bit more TBH

dwalluck:
> This was changed in #246
> 
> This does change the logging and how exceptions are propagated. I think I'd like to understand the difference with 4.10 a bit more TBH

Wasn't it added in order to produce "Build exception but unable to transfer message due to mix of JDK versions." That check is still there in this PR, but Gradle 4.10 won't hit it because it throws a different exception than the one it is looking for.

It is suffers this same issue still, then this check needs to be updated as well, I think, but that is related to `DifferentJVMTest` and not `MainTest`. For `MainTest`, we know the code still works since we properly get "must be configured in order for dependency scanning to work", but we can't check `e.getCause().getMessage()`, since the exception chains are different, but checking the log itself, we know it got there eventually.

I think it would help to post the stack traces of 4.10 vs. 5 here.

rnc:
@dwalluck I think this can be closed as it was replaced by #318  ?

PR #314, opened 2021-07-08
NCL-6050: Skip certain tests based on current Gradle version

dwalluck:


rnc:
While this is obviously transparent to current tests, would it not be better as part of a set of commits for the 4.10 testing? Merging this without that doesn't seem to make sense?

dwalluck:
> While this is obviously transparent to current tests, would it not be better as part of a set of commits for the 4.10 testing? Merging this without that doesn't seem to make sense?

It could. The combined PR is currently coming in at 45 files, whereas this is only 16.

I will need to update the tests here anyway. The PR will actually cover 4.10 as well as some older versions of 5.x (because I actually tried to figure out exactly which 5.x version certain features were added in).

Whether we should add these 5.x versions to the GitHub CI, I am not sure. At least we don't need ALL of 5.0, 1, 2, 3, 4, 5 test. But, based on the tests, it could have 0, 2, and 4, I think.

dwalluck:
I will close this and open a new PR.

PR #315, opened 2021-07-12
Make --target default to ${user.dir}

dwalluck:


cli/src/main/java/org/jboss/gm/cli/Main.java
```diff
@@ -65,7 +65,7 @@
     @Option(names = "--trace", description = "Enable trace.")
     private boolean trace;
 
-    @Option(names = { "-t", "--target" }, required = true, description = "Target Gradle directory.")
+    @Option(names = { "-t", "--target" }, required = true, description = "Target Gradle directory.", defaultValue = ".")
```
rnc, line 68: I wonder if we should match it to PME https://github.com/release-engineering/pom-manipulation-ext/blob/master/cli/src/main/java/org/commonjava/maven/ext/cli/Cli.java#L109 (and update the docs of course)

dwalluck: I switched from "." to "${user.dir} since PME uses that.

dwalluck: Added to help output "default: ${DEFAULT-VALUE}", but I kind of don't like it as that makes it "dynamic".

dwalluck: I checked the gh-pages branch. Aside from maybe changing the help output, there is really no mention of `--target`.

cli/src/main/java/org/jboss/gm/cli/Main.java
```diff
@@ -65,7 +65,9 @@
     @Option(names = "--trace", description = "Enable trace.")
     private boolean trace;
 
-    @Option(names = { "-t", "--target" }, required = true, description = "Target Gradle directory.")
+    @Option(names = { "-t", "--target" },
+            description = "Target Gradle directory (default: ${DEFAULT-VALUE}).",
+            defaultValue = "${sys:user.dir}")
```
rnc, line 70: Nice!

PR #318, opened 2021-07-16
NCL-6050: Add support for Gradle 4.10

dwalluck:


analyzer/src/functTest/java/org/jboss/gm/analyzer/alignment/InvalidProjectFunctionalTest.java
```diff
@@ -52,8 +52,8 @@ public void ensureInvalidException() throws IOException, URISyntaxException {
             TestUtils.align(projectRoot, projectRoot.getName(), true);
             fail("No exception thrown");
         } catch (ManipulationUncheckedException e) {
-            assertThat(e.getMessage().contains(
-                    "For configuration compile; unable to resolve all dependencies: [io.undertow:undertow-core:2.0+, org.apache.commons:commons-lang3:3.8.1.redhat-3]"));
+            assertThat(e.getMessage()).contains(
```
rnc, line 55: Good catch!

analyzer/src/functTest/java/org/jboss/gm/analyzer/alignment/DynamicWithLocksProjectFunctionalTest.java
```diff
@@ -94,7 +98,7 @@ public void ensureAlignmentFileCreated()
                                 tuple("resteasy-jaxrs", "3.6.3.SP1-redhat-00010"),
                                 tuple("HdrHistogram", "2.1.10"));
 
-                assertThat(root.getAlignedDependencies().keySet()).containsOnly(
+                assertThat(root.getAlignedDependencies()).containsOnlyKeys(
```
rnc, line 101: Nice

build.gradle.kts
```diff
@@ -349,31 +445,36 @@ fun MavenPublication.generatePom() {
...
                 id.set("TomasHofman")
                 name.set("Tomas Hofman")
                 email.set("thofman@redhat.com")
             }
+
             developer {
                 id.set("metacosm")
                 name.set("Chris Laprun")
                 email.set("claprun@redhat.com")
             }
```
rnc, line 475: Add your name?

cli/src/test/java/org/jboss/gm/cli/DifferentJVMTest.java
```diff
@@ -49,38 +58,67 @@
...
+    public static void setupJVM() throws IOException {
+        assumeTrue(SystemUtils.IS_OS_LINUX);
+
+        String filename = null;
+        UnArchiver ua = null;
+
+        if (SystemUtils.IS_OS_LINUX) {
+            filename = "OpenJDK8U-jdk_x64_linux_hotspot_8u272b10.tar.gz";
+            ua = new TarGZipUnArchiver();
+        } else if (SystemUtils.IS_OS_WINDOWS) {
```
rnc, line 85: Why are we testing for Windows if this is only run on Linux?

.github/workflows/gradle.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        gradle: [5.6.4, 6.4.1, 6.5.1, 6.6.1, 6.7.1, 6.8.3, 6.9]
+        gradle: [4.10.3, 5.0, 5.1.1, 5.2.1, 5.3.1, 5.4.1, 5.5.1, 5.6.4, 6.0.1, 6.1.1, 6.2.2, 6.3, 6.4.1, 6.5.1, 6.6.1, 6.7.1, 6.8.3, 6.9]
```
rnc, line 17: >> @rnc 
>> Wow!

> @dwalluck 
> We could remove these in the future if it's too excessive, although I'd love to see at least one run pass.

No, its fine, but kind of impressive!

cli/src/test/java/org/jboss/gm/cli/DifferentJVMTest.java
```diff
@@ -49,38 +58,67 @@
...
 
-            compressedFile.deleteOnExit();
+            System.out.println("Set hook!");
+            Runtime.getRuntime().addShutdownHook(new Thread(() -> {
+                try {
+                    Files.deleteIfExists(destFile);
+                } catch (IOException e) {
+
+                }
+            }));
```
rnc, line 121: >> @rnc 
>> Why the change to use shutdown hook?

> @dwalluck 
> Only because I switched to Path which didn't have that API directly, but I can change this back.

Yeah I think so especially as we are only running on Linux

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-5.6.4-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.10.3-all.zip
```
rnc, line 3: >> @rnc 
> > Lets revert this please - I don't think we need to change the wrapper ? Or this a work in progress thing while you debug and examine APIs inside the IDE?

> @dwalluck 
> I just changed it while working on 4.10 support. I did not mean for it to be permanent.

Ok no problem!

build.gradle.kts
```diff
@@ -6,43 +6,80 @@ import org.gradle.util.GradleVersion
...
+        id("io.freefair.lombok") version "5.3.3.3" apply false
     }
-}
 
-apply(plugin = "net.researchgate.release")
+    if (org.gradle.util.GradleVersion.current() >= org.gradle.util.GradleVersion.version("6.0")) {
+        id("org.kordamp.gradle.jacoco") version "0.46.0"
+    } else if (org.gradle.util.GradleVersion.current() >= org.gradle.util.GradleVersion.version("5.2")) {
+        id("org.kordamp.gradle.jacoco") version "0.30.0"
+    }
```
rnc, line 57: >> @rnc 
>> Any particular reason its using a switch (never seen when in Kotlin before!) and a if-block ?

> @dwalluck 
> I am not sure what you're saying the exact problem is.
> 
> The IDE suggested changing a long series of if-s to when, but what I really need is some construct with fall-though since the first block is for versions < 5.0 and the if is for versions >= 5.2 and I can't seem to handle a case like this in an if since it's mutually exclusive without duplicating everything.

Sorry re-read the logic and its testing different versions for different plugins ; I see your point.

dwalluck:
> Why the empty `lombok.config` file?

There's a link to the issue in the build file. It's a bug in older versions of the lombok plugin. If this file doesn't exist it seems to write its own and fail (it ignores the fact that we've disabled these files for future versions of the plugin).

common/src/main/java/org/jboss/gm/common/Configuration.java
```diff
@@ -57,9 +57,11 @@ default Properties getProperties() {
     }
 
     @Key(DA)
+    @DefaultValue("")
     String daEndpoint();
 
     @Key("AProxDeployUrl")
+    @DefaultValue("")
```
rnc, line 64: >> @rnc 
>> Do we really need to default to empty String as opposed to null ?

> @dwalluck 
> Something weird is going on with 4.10. What generates this code, is it lombok? But forcing them to "" makes the value consistent across all versions. I wonder if all Strings are exhibiting this issue though.
> 
> To test, remove these lines, and remove the change to analyzer/src/main/java/org/jboss/gm/analyzer/alignment/DAAlignmentService.java and you should see a test failure.

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/DAAlignmentService.java
```diff
@@ -37,7 +38,7 @@ public DAAlignmentService(Configuration configuration) {
 
         final String endpointUrl = configuration.daEndpoint();
 
-        if (endpointUrl == null && dependencySource != NONE) {
+        if (isEmpty(endpointUrl) && dependencySource != NONE) {
```
rnc, line 41: > @dwalluck There is something weird going on between versions. With 4.10, the default value seems to be defaulting to "" and not null and so this check fails, which is why I changed the URL default to "".

But if you've changed this to `isEmpty` can't we revert the configuration `DefaultValue` changes?

dwalluck: I think we can revert.

dwalluck:
> Why the empty `lombok.config` file?

`generateLombokConfig.enabled = false` seems to fail in older versions. A workaround was to create empty files, then it will not try to create new ones.

.github/workflows/gradle.yml
```diff
@@ -14,7 +14,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        gradle: [5.6.4, 6.4.1, 6.5.1, 6.6.1, 6.7.1, 6.8.3, 6.9]
+        # XXX: Remove 5.0 as it becomes 5 (eskatos/gradle-command-action bug?)
+        gradle: [4.10.3, 5.1.1, 5.2.1, 5.3.1, 5.4.1, 5.5.1, 5.6.4, 6.0.1, 6.1.1, 6.2.2, 6.3, 6.4.1, 6.5.1, 6.6.1, 6.7.1, 6.8.3, 6.9]
```
dwalluck, line 18: I've removed 5.0. It says it can't find version 5.

dwalluck: It needed to be quoted since 5.0 == 5.

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/DAAlignmentService.java
```diff
@@ -37,7 +38,9 @@ public DAAlignmentService(Configuration configuration) {
 
         final String endpointUrl = configuration.daEndpoint();
 
-        if (endpointUrl == null && dependencySource != NONE) {
+        logger.info("endpointUrl = {}, dependencySource = {}", endpointUrl, dependencySource);
```
rnc, line 41: Perhaps debug level logging?

build.gradle.kts
```diff
@@ -6,43 +6,82 @@ import org.gradle.util.GradleVersion
...
+
+    if (org.gradle.util.GradleVersion.current() >= org.gradle.util.GradleVersion.version("5.3")) {
+        id("org.kordamp.gradle.jacoco") version "0.46.0"
     }
 }
 
-apply(plugin = "net.researchgate.release")
+if (org.gradle.util.GradleVersion.current() < org.gradle.util.GradleVersion.version("5.3")) {
+    tasks.register("AggregateJacocoReport");
```
rnc, line 59: Neat! Wonder what happens if there is no report when codecov task runs in GitHub? .... I presume it just ignores it.

Also, could we add a comment line as to why as I think this is something that isn't obvious at first glance

cli/src/main/java/org/jboss/gm/cli/Main.java
```diff
@@ -219,19 +218,16 @@ private void executeGradle() throws ManipulationException {
...
-            if (e.getCause().getClass().getName().equals("org.gradle.api.UncheckedIOException")) {
+            if (e.getCause() != null && "org.gradle.api.UncheckedIOException".equals(e.getCause().getClass().getName())) {
                 logger.debug("Hit https://github.com/gradle/gradle/issues/9339 ", e);
                 logger.error(
                         "Build exception but unable to transfer message due to mix of JDK versions. Examine log for problems");
             } else {
                 logger.error("Gradle connection exception", e);
             }
-            throw new ManipulationException("Problem executing build");
+            throw new ManipulationException("Problem executing build", e);
```
rnc, line 230: Marking this section for discussion (so I don't forget)

cli/src/test/java/org/jboss/gm/cli/MainTest.java
```diff
@@ -129,7 +132,7 @@ public void testInvokeGroovy() throws Exception {
         m.run(args);
 
         assertTrue(systemOutRule.getLog().contains("Running Groovy script on"));
-        assertTrue(systemOutRule.getLog().contains("Tasks runnable from root project"));
+        assertTrue(systemOutRule.getLog().contains("asks runnable from root project"));
```
rnc, line 135: Out of curiosity - did prior versions not capitalize?

cli/src/test/java/org/jboss/gm/cli/MainTest.java
```diff
@@ -211,6 +215,8 @@ public void testInvokeAlignment() throws Exception {
 
     @Test
     public void testInvokeAlignmentFails() throws Exception {
+        assumeTrue(!SystemUtils.IS_OS_WINDOWS);
```
rnc, line 218: We've got a right mix of asserting platforms ... I think we need to discuss whether - we need this, whether it can be simplified etc.

cli/src/test/java/org/jboss/gm/cli/MainTest.java
```diff
@@ -249,14 +255,19 @@ public void testInvokeAlignmentFails() throws Exception {
...
+            // ignored as exception can vary between gradle versions and systemErr should contain correct message
         }
-        assertTrue(systemErrRule.getLog().contains("'restURL' must be configured in order for dependency scanning to work"));
+
         assertFalse(systemErrRule.getLog().contains(ANSIConstants.ESC_START));
+        // NCL-6050: FIXME
+        // DAAlignmentService.java:51 No dependencySource configured ; unable to call endpoint?
+        assumeTrue(GradleVersion.current().compareTo(GradleVersion.version("5.0")) >= 0);
+        assertTrue(systemErrRule.getLog()
+                .contains("'" + Configuration.DA + "' must be configured in order for dependency scanning to work"));
```
rnc, line 270: Marker comment to discuss the `FIXME`

cli/build.gradle.kts
```diff
@@ -26,17 +27,24 @@ dependencies {
...
     "jar"(Jar::class) {
-        this.manifest {
+        manifest {
             attributes["Main-Class"] = "org.jboss.gm.cli.Main"
         }
     }
+
+    "test" {
+        dependsOn(":analyzer:shadowJar")
+    }
```
rnc, line 49: Why this change please?

build.gradle.kts
dwalluck: I wonder how it handles multiple reports in general, since we want it to use the one with the highest coverage.

cli/build.gradle.kts
dwalluck: This actually came from a complaint from Gradle 7, I believe. We could remove it until that support is complete. It complains that the `cli` depends on the `shadowJar` task, but doesn't declare it anywhere. There is `testImplementation(project(":analyzer"))` but that doesn't seem to be doing it.

cli/src/test/java/org/jboss/gm/cli/MainTest.java
dwalluck: I believe that was it. I had wanted to use something like `.matches("[Tt]")` instead of `.contains()`, but out of curiosity, do we need these first two lines at all? The next three are what actually test out output, whereas these two are from Gradle itself.

build.gradle.kts
rnc: Yeah I was wondering that - hehe originally thinking of running coverage for only one version. Unless it accumulates them?

cli/src/test/java/org/jboss/gm/cli/MainTest.java
rnc: 134 looks like ours, but yes you are right 135 might not be needed

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/DAAlignmentService.java
dwalluck: I tried info, debug, and println and I am not seeing it in the log. It's like it's not being called.

build.gradle.kts
dwalluck: I thought it aggregates them. At the very least, you can sometimes see the results change before your eyes if you watch a PR where the runs are in progress.

cli/src/test/java/org/jboss/gm/cli/MainTest.java
dwalluck: May just be better to remove this one line then.

cli/src/test/java/org/jboss/gm/cli/MainTest.java
dwalluck: I am removing the assertions from this file.

cli/src/test/java/org/jboss/gm/cli/DifferentJVMTest.java
dwalluck: I was actually trying to see if I could make it more generic, like if I had a `.zip` file and didn't have a `.tar.gz`. In the end, I found that I actually ran into a different issue.

1. The archive failed to extract, due to space or I don't know, but it creates the directory first, so the check for the directory passed, but there was no JDK in there. I slightly changed it to check for `bin/` instead, but that's not foolproof either.
2. I do not like that a test extracts a permanent JDK into my home directory, but this is a different issue that could perhaps be improved.
3. At least if we're extract to a known location, we could cache this in the CI so it doesn't need to download it each time.

rnc: 2/3) We should be caching in the CI ( I hope) ; as regards the local cache of a JDK it should be replicating the same as https://docs.gradle.org/current/userguide/toolchains.html

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/DAAlignmentService.java
rnc: I am seeing it if I run manually ; remember Gradle has lifecycle logging as well which is the default

cli/src/test/java/org/jboss/gm/cli/MainTest.java
rnc: But isn't this line verifying that Gradle has been invoked correctly?

cli/build.gradle.kts
rnc: Ok ; see https://github.com/project-ncl/gradle-manipulator/pull/319 which changes it slightly.

cli/src/test/java/org/jboss/gm/cli/DifferentJVMTest.java
dwalluck: I meant in the GitHub CI where we cache other build-related files between runs.

build.gradle.kts
```diff
@@ -180,26 +229,75 @@ subprojects {
...
+        classifier = "test-fixtures"
+        from(sourceSets["testFixtures"].output)
+    }
+
+    if (project.name == "common") {
+        artifacts {
+            add("archives", tasks["testFixturesJar"])
+            add("testFixturesCompile", testFixturesJar)
+        }
+    }
```
rnc, line 289: Couple of questions / comments:
1. I guess this is simulating the test-fixture plugin from 5.6. Does it install / publish any form of jar ? 
2. Should it use the original fixtures for >=5.6 ? (I'm inclined to say no especially if this publishes an extra jar with a classifier? ) 

Can we a couple of sentences at the top explaining why we are using this block that looks like 5.6 test fixtures?

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/DAAlignmentService.java
dwalluck: I was checking in the HTML produced after a build failure (the FIXME in `MainTest`). I do see the line from another method "DAAlignmentService.java:51 No dependencySource configured ; unable to call endpoint?", but not this one.

That line implies `dependencySource == NONE` which won't trigger the `if` which has `dependencySource != NONE`, although the test seems to expect it to trigger.

build.gradle.kts
```diff
@@ -212,16 +310,20 @@ subprojects {
...
+            if (org.gradle.util.GradleVersion.current() >= org.gradle.util.GradleVersion.version("5.0")) {
+                // Minimise the resulting uber-jars to ensure we don't have massive jars
+                minimize {
+                    // Sometimes minimisation takes away too much ... ensure we keep these.
+                    exclude(dependency("com.fasterxml.jackson.core:.*:.*"))
+                    exclude(dependency("org.commonjava.maven.ext:.*:.*"))
+                    exclude(dependency("org.commonjava.maven.atlas:.*:.*"))
+                    exclude(dependency("org.aeonbits.owner:.*:.*"))
+                    exclude(dependency("org.slf4j:.*:.*"))
+                    exclude(dependency("org.apache.maven:.*:.*"))
```
rnc, line 323: Good choice, keeps it simple. I would however modify the comment to explain why we only minimise on >= 5.0

cli/src/test/java/org/jboss/gm/cli/DifferentJVMTest.java
rnc: +1 , I meant that and while I thought it was, perhaps it isn't from looking at it.

build.gradle.kts
rnc: Yeah I think it does merge them. 
Can we get a comment line to explain why an empty task has been added to resolve this?

cli/src/test/java/org/jboss/gm/cli/MainTest.java
rnc: Huh, good question!

build.gradle.kts
```diff
@@ -6,43 +6,87 @@ import org.gradle.util.GradleVersion
...
 
-    } else {
-        id("com.github.johnrengelman.shadow") version "6.1.0"
+    when {
+        org.gradle.util.GradleVersion.current() < org.gradle.util.GradleVersion.version("5.0") -> {
+            id("io.freefair.lombok") version "2.9.5" apply false
+        }
+        org.gradle.util.GradleVersion.current() < org.gradle.util.GradleVersion.version("5.2") -> {
+            // XXX: Versions 3.x < 3.6.1 suffer from <https://github.com/freefair/gradle-plugins/issues/31>
+            id("io.freefair.lombok") version "3.0.0" apply false
```
rnc, line 48: So this is what causes the lombok.config files - for Gradle 5.0 -> 5.1 ? What about if we only set that ` generateLombokConfig.enabled = false` on appropriate versions ? And ignore any generated files?

dwalluck: > So this is what causes the lombok.config files - for Gradle 5.0 -> 5.1 ?

It's Gradle < 5.2 apparently, so 4.10 is included as well.

> What about if we only set that ` generateLombokConfig.enabled = false` on appropriate versions ? And ignore any generated files?

We already disable it on all versions so this applies to 4.10, 5.0, and 5.1 as well.

However, that plugin version ignores this setting and there seems to be no way to turn it off. Creating empty files is a way to work around it creating (invalid) files since the setting does not have any effect.

If there's some way to set it to a file in the root project, then at least we could cut it down to a single file instead of one per module.

build.gradle.kts
dwalluck: I added a comment. Also, nothing else seemed to work for me except to not run it at all (meaning `exclude()` had no effect).

cli/src/test/java/org/jboss/gm/cli/DifferentJVMTest.java
dwalluck: I think the Gradle cache is something separate. I am not sure that it would cache the JDK itself.

For example, for Maven builds we might cache `.m2/repository` or anything else we download.

build.gradle.kts
dwalluck: Added comment.

cli/src/test/java/org/jboss/gm/cli/MainTest.java
dwalluck: Right now I'm removed the `"asks"` line. Let me know if I should put it back. If we do, using `assertThat().matches()` may be clearer.

build.gradle.kts
dwalluck: > 1. I guess this is simulating the test-fixture plugin from 5.6. Does it install / publish any form of jar ?

Yes, common test-fixtures is published before, so I was trying to also publish this one.

> 2. Should it use the original fixtures for >=5.6 ? (I'm inclined to say no especially if this publishes an extra jar with a classifier? )

I don't think we can, in that it's not something we can just `if` out on older versions.

> Can we a couple of sentences at the top explaining why we are using this block that looks like 5.6 test fixtures?

Will do.

cli/src/main/java/org/jboss/gm/cli/Main.java
dwalluck: I'm reverting as much as I can here.

The issue is we have three different possible code paths/messages here depending on the Gradle version so it's not as simple as just checking the exception type or message---the code still works (in the sense that an exception is thrown here), but it's hard to test, especially since we're printing slightly different messages in each case.

cli/src/test/java/org/jboss/gm/cli/MainTest.java
dwalluck: ?[90m[DAAlignmentService.java:41] ?[0mendpointUrl = null, dependencySource = NONE

But, the default is supposed to be `dependencySource = REST` and it doesn't look like it's been set anywhere.

dwalluck: I thought it was something to do with not processing the annotations properly.

cli/src/test/java/org/jboss/gm/cli/DifferentJVMTest.java
dwalluck: But, Gradle cache may be a good idea nonetheless. I worry, though, as we seem to have "corrupted" artifacts in our caches that we had to manually remove.

analyzer/build.gradle.kts
```diff
@@ -45,8 +46,9 @@ dependencies {
 
     testRuntimeOnly("commons-io:commons-io:${project.extra.get("commonsVersion")}")
     testImplementation("org.commonjava.maven.ext:pom-manipulation-common:${project.extra.get("pmeVersion")}")
-    testImplementation(testFixtures(project(":common")))
-    testImplementation(project(":common"))
+    //testImplementation(project(":common"))
+    //testImplementation(testFixtures(project(":common")))
```
rnc, line 50: Commented out code.

build.gradle.kts
rnc: I have tested on both 4.10 and 5.1 adding ` if (org.gradle.util.GradleVersion.current() >= org.gradle.util.GradleVersion.version("5.2")) { ` around the lombok configuration block ; with that, even with no lombok files it seems to work. It does generate them but I think we could ignore that?

build.gradle.kts
```diff
@@ -308,21 +413,22 @@ subprojects {
     }
 
     tasks.withType<JavaCompile>().configureEach {
+        options.compilerArgs.add("-Xlint:deprecation")
```
rnc, line 416: Any particular reason to add this?

cli/build.gradle.kts
```diff
@@ -26,21 +27,24 @@ dependencies {
     runtimeOnly("org.apache.maven:maven-artifact:${project.extra.get("mavenVersion")}")
 
     testRuntimeOnly("commons-io:commons-io:${project.extra.get("commonsVersion")}")
-    testImplementation(testFixtures(project(":common")))
+    //testImplementation(testFixtures(project(":common")))
```
rnc, line 30: Commented out line

common/build.gradle.kts
```diff
@@ -34,8 +33,8 @@ dependencies {
     // This is to prevent compilation errors in conjunction with Lombok due to use of PME code.
     compileOnly("org.apache.maven:maven-compat:3.5.0")
 
-    testFixturesImplementation(gradleApi())
-    testFixturesImplementation("junit:junit:${project.extra.get("junitVersion")}")
+    //testFixturesImplementation(gradleApi())
+    //testFixturesImplementation("junit:junit:${project.extra.get("junitVersion")}")
```
rnc, line 37: Commented out lines

cli/src/main/java/org/jboss/gm/cli/Main.java
rnc: I've been investigating the different exceptions and the failing 4.10 test with restURL ; its subtle to do with prior tests (runAlignment) influencing latter due to cached objects when the plugin is run within the daemon ... I think - essentially the ConfigCache _within_ the AlignmentTask is carrying over which is interfering.

build.gradle.kts
dwalluck: I was wondering what it might show in relation to Gradle deprecration itself but I don't think it does. I will remove it.

build.gradle.kts
dwalluck: I have also added `lombok.config` to `.gitignore`. We never generated the file in the past, but as some versions generate the files now, it's safer.

manipulation/src/functTest/resources/kafka/build.gradle
```diff
@@ -622,10 +622,11 @@ task jacocoRootReport(type: org.gradle.testing.jacoco.tasks.JacocoReport) {
...
 
-  additionalSourceDirs.from = javaProjects.sourceSets.main.allSource.srcDirs
-  sourceDirectories.from = javaProjects.sourceSets.main.allSource.srcDirs
-  classDirectories.from = javaProjects.sourceSets.main.output
-  executionData.from = javaProjects.jacocoTestReport.executionData
+  // XXX: Commented out due to incompatibility with Gradle 4.10
+  //additionalSourceDirs.from = javaProjects.sourceSets.main.allSource.srcDirs
+  //sourceDirectories.from = javaProjects.sourceSets.main.allSource.srcDirs
+  //classDirectories.from = javaProjects.sourceSets.main.output
+  //executionData.from = javaProjects.jacocoTestReport.executionData
```
dwalluck, line 629: Not sure why these don't work; tests seem to pass without them.

build.gradle.kts
```diff
@@ -142,10 +191,15 @@ subprojects {
     apply(plugin = "com.adarshr.test-logger")
     apply(plugin = "io.freefair.lombok")
 
-    extra["lombokVersion"] = extensions.findByType(LombokExtension::class)?.version?.get()
-    tasks.named("generateLombokConfig") {
-        // Don't generate lombok.config files ( https://docs.freefair.io/gradle-plugins/3.6.6/reference/#_lombok_config_handling )
-        enabled = false
+    extra["lombokVersion"] = extensions.findByType(LombokExtension::class)?.version
```
dwalluck, line 194: I am not sure if the `?` is strictly necessary here. I mean, we load the Lombok extension on all Gradle versions (we can't build without it).

I had changed it from `.version?.get()` to `version` for older versions.

build.gradle.kts
rnc: Thats perfect, thanks.

build.gradle.kts
rnc: Ok cool

manipulation/src/functTest/resources/kafka/build.gradle
rnc: I think that test was introduced for NCLSUP175 (Handle archivesBaseName for subprojects) so removing those should be fine.

cli/src/test/java/org/jboss/gm/cli/DifferentJVMTest.java
rnc: Let move the cache / windows discussion to a separate ticket (NCL-6679) to move this forward

rnc (approved):

PR #320, opened 2021-07-30
NCL-6544: Add support for Gradle 7

dwalluck:


analyzer/src/functTest/java/org/jboss/gm/analyzer/alignment/InvalidProjectFunctionalTest.java
```diff
@@ -52,8 +54,16 @@ public void ensureInvalidException() throws IOException, URISyntaxException {
...
+            String regex = ".*For configuration compileClasspath, unable to resolve all project dependencies: \\[(.*)].*";
+            Pattern pattern = Pattern.compile(regex, Pattern.DOTALL);
+            Matcher matcher = pattern.matcher(e.getMessage());
+            assertThat(matcher.matches()).isTrue();
+            assertThat(matcher.groupCount()).isEqualTo(1);
+            String[] dependencies = matcher.group(1).split(", ");
+            String[] expectedDependencies = { "io.undertow:undertow-core:2.0+",
+                    "org.apache.commons:commons-lang3:3.8.1.redhat-3",
+                    "org.springframework:spring-context:5.1.6.INVALID_VERSION" };
+            assertThat(dependencies).containsExactly(expectedDependencies);
```
rnc, line 66: Can you explain why the test check is now different please?

analyzer/src/functTest/java/org/jboss/gm/analyzer/alignment/SimpleProjectFunctionalTest.java
```diff
@@ -322,12 +322,12 @@ public void verifyAlignmentReportJson() throws IOException, URISyntaxException {
...
                         "\tProject version : 1.0.1 --> 1.0.1.redhat-00002%n" +
                         "%n" +
-                        "\tDependencies : com.yammer.metrics:metrics-core:2.2.0 --> com.yammer.metrics:metrics-core:2.2.0-redhat-00001%n"
-                        +
                         "\tDependencies : org.hibernate:hibernate-core:5.3.7.Final --> org.hibernate:hibernate-core:5.3.7.Final-redhat-00001%n"
                         +
                         "\tDependencies : io.undertow:undertow-core:2.0.15.Final --> io.undertow:undertow-core:2.0.15.Final-redhat-00001%n"
                         +
+                        "\tDependencies : com.yammer.metrics:metrics-core:2.2.0 --> com.yammer.metrics:metrics-core:2.2.0-redhat-00001%n"
+                        +
```
rnc, line 330: Is this change due to it being sorted now?

analyzer/src/functTest/resources/simple-existing-manipulation/build.gradle
```diff
@@ -12,8 +12,8 @@ repositories {
...
 
 dependencies {
-    compile group: 'org.apache.commons', name: 'commons-lang3', version: '3.8.1'
-    compile group: 'org.hibernate', name: 'hibernate-core', version: '5.3.7.Final'
-    compile group: 'io.undertow', name: 'undertow-core', version: '2.0.15.Final'
-    testCompile group: 'junit', name: 'junit', version: '4.12'
+    implementation group: 'org.apache.commons', name: 'commons-lang3', version: '3.8.1'
+    implementation group: 'org.hibernate', name: 'hibernate-core', version: '5.3.7.Final'
+    implementation group: 'io.undertow', name: 'undertow-core', version: '2.0.15.Final'
+    testImplementation group: 'junit', name: 'junit', version: '4.12'
```
rnc, line 18: This is because compile/testCompile were removed from 7 ; is that correct?

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
```diff
@@ -331,6 +329,7 @@ private void align(Configuration configuration, ManipulationCache cache, Manipul
         writeGmeConfigMarkerFile(rootProject.getBuildFile());
         writeGmeReposMarkerFile();
         writeRepositorySettingsFile(cache.getRepositories());
+        final Set<ProjectVersionRef> nonAligned = new LinkedHashSet<>();
```
rnc, line 332: You're wanting to maintain order?

cli/build.gradle.kts
```diff
@@ -44,6 +44,11 @@ tasks {
     }
 
     "test" {
+        dependsOn("shadowJar")
         dependsOn(":analyzer:assemble")
     }
+
+    "publishShadowPublicationToMavenLocal" {
+        dependsOn("jar")
+    }
```
rnc, line 53: Can you clarify these please?

manipulation/build.gradle.kts
```diff
@@ -90,3 +92,25 @@ val functionalTest = task<Test>("functionalTest") {
...
+    dependsOn("shadowJar")
+}
+
+tasks.getByName("functionalTest") {
+    dependsOn("shadowJar")
+}
+
+tasks.getByName("publishShadowPublicationToMavenLocal") {
+    dependsOn("publishPluginJavaDocsJar", "publishPluginJar", "jar")
+}
```
rnc, line 116: I think I'd like to understand why this is needed as well.

manipulation/src/main/java/org/jboss/gm/manipulation/actions/UploadTaskTransformerAction.java
```diff
@@ -1,40 +1,96 @@
...
         this.alignmentConfiguration = alignmentConfiguration;
         this.resolvedDependenciesRepository = resolvedDependenciesRepository;
+
+        try {
+            resolverClass = Class.forName("org.gradle.api.artifacts.maven.MavenResolver")
+                    .asSubclass(ArtifactRepository.class);
+            getPomMethod = resolverClass.getMethod("getPom");
+            // XXX: This class is in Gradle 7, but deprecated
+            // XXX: See <https://docs.gradle.org/current/userguide/upgrading_version_7.html#upload_task_deprecation>
+            uploadClass = Class.forName("org.gradle.api.tasks.Upload").asSubclass(Task.class);
```
rnc, line 60: I'm slightly confused by this comment - if this class _is_ in Gradle 7 why is reflection required?

analyzer/src/functTest/java/org/jboss/gm/analyzer/alignment/InvalidProjectFunctionalTest.java
dwalluck: 1. There's the addition of "org.springframework:spring-context:5.1.6.INVALID_VERSION" which was always there, but didn't show up until we switched all to `implementation` (for Gradle 7). This shows now as configuration `compileClasspath` instead of `compile`.

2. There was a problem with the ordering when switching between Gradle 7 and other versions. With `HashSet`, we don't seem to get a guaranteed order (even if it usually works, since it's not "random"). This code was changed originally to be able to handle changes in order by using `contains` instead of a simple string comparison. However, since I changed certain spots to maintain the original set order, this change should no longer be necessary (but, I still found it to be more flexible).

analyzer/src/functTest/java/org/jboss/gm/analyzer/alignment/SimpleProjectFunctionalTest.java
dwalluck: Yes, the order is not sorted, but _maintained_ from the original order.

Here, original usually means as returned from a Gradle API. We'd then lose this order by putting the items into a `HashSet`.

I didn't try to change everywhere, only cases where we printed or checked the order, so that the order should be predictable now.

manipulation/src/main/java/org/jboss/gm/manipulation/actions/UploadTaskTransformerAction.java
dwalluck: I put a link to the manual which says that this class is deprecated and will be removed in Gradle 8. As it was using reflection for the other classes, it was just a bit of futureproofing.

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
dwalluck: Since it maintains order for the other printed dependencies (as caught by a test), then it makes sense not to lose order for the non-aligned as well. Because we print these and/or verify their order, they should be predictable and `HashSet` does not guarantee order.

analyzer/src/functTest/resources/simple-existing-manipulation/build.gradle
dwalluck: Yes, so at the very least, every test that is compatible with Gradle 7 needs to be changed.

I also tried to change as many build files as I could although sometimes "read-only" configuration is now an issue (I put a note). In order words, you could directly modify the `compile` configuration, but not the new ones.

cli/build.gradle.kts
dwalluck: Granted, for this task, it only happens when you use `publishToMavenLocal`. But, in general:

There's a new warning output when built with Gradle 7+. It is explained in the following links
<https://docs.gradle.org/7.0/release-notes.html#build-reliability-improvements> and <https://docs.gradle.org/7.0/userguide/validation_problems.html#implicit_dependency>.

It is of the form:

Execution optimizations have been disabled for task ':task' to ensure correctness due to the following reasons:
  - Gradle detected a problem with the following location: 'location'. Reason: Task ':task' uses this output of task ':otherTask' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed. Please refer to https://docs.gradle.org/7.0/userguide/validation_problems.html#implicit_dependency for more details about this problem.

manipulation/src/main/java/org/jboss/gm/manipulation/actions/UploadTaskTransformerAction.java
```diff
@@ -1,40 +1,96 @@
...
+    private Method getRepositoriesMethod;
+    private Method withXmlMethod;
 
     public UploadTaskTransformerAction(ManipulationModel alignmentConfiguration,
             ResolvedDependenciesRepository resolvedDependenciesRepository) {
+        disabled = GradleVersion.current().compareTo(GradleVersion.version("7.0")) >= 0;
+
+        if (disabled) {
+            return;
+        }
```
rnc, line 49: Out of curiosity, if a build does configure a UploadTask action in Gradle 7, can it ever even get to this point?

dwalluck: It was more a guard for the Java code itself as I am not entirely sure when or how the class can be created.

Generally, in a build file itself, it would fail as soon as an attempt is made to use the maven plugin, before the alignment can even be run.

Also, looking at <https://github.com/project-ncl/gradle-manipulator/blob/4fa608d33161af3d24b354fd1a2948518184ee66/manipulation/src/main/java/org/jboss/gm/manipulation/ManipulationPlugin.java#L156>, why wouldn't we move the check from <https://github.com/project-ncl/gradle-manipulator/blob/4fa608d33161af3d24b354fd1a2948518184ee66/manipulation/src/main/java/org/jboss/gm/manipulation/actions/UploadTaskTransformerAction.java#L30>to there.

Can `execute` be called from somewhere else?

In other words, more generally, can we remove any guards from `UploadTaskTransformerAction` and move it to `ManipulationPlugin`? Or is there some other possible way to call this?

analyzer/src/functTest/resources/simple-existing-manipulation/build.gradle
rnc: Thanks.

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
rnc: Ok, thanks.

analyzer/src/functTest/java/org/jboss/gm/analyzer/alignment/SimpleProjectFunctionalTest.java
rnc: Ok sure.

dwalluck: I know  we are currently using a `TreeMap` to sort something alphabetically. I think it's the project names.

If you'd rather sort all output alphabetically, I think I'd still rather store internally in Gradle's order (which I think is in encountered order, like a `List` would do), then use `sorted()` when we go to print it. Although, I am not proposing to change it in this PR.

manipulation/build.gradle.kts
dwalluck: This happens when using `publishToMavenLocal`. I assume `publish` would have the same issues, but I can't exactly test a real publish. We only `publish` officially under Gradle 5.6, but I think we might `publishToMavenLocal` for any version.

manipulation/src/main/java/org/jboss/gm/manipulation/actions/UploadTaskTransformerAction.java
dwalluck: I guess it affects `LegacyMavenPublishingRepositoryAction` also, which uses this class directly. But, I guess you're saying not to worry about it until Gradle 8.

rnc: Ok thanks, that makes sense.

manipulation/src/main/java/org/jboss/gm/manipulation/actions/UploadTaskTransformerAction.java
rnc: Yeah I agree, I think the original guards were paranoia but whether we _really_ need them is another question. I don't think it should be possible to call those classes ... as per my comment line in one of the classes `This should never happen due to prior checks in ManipulationPlugin`

analyzer/src/functTest/java/org/jboss/gm/analyzer/alignment/InvalidProjectFunctionalTest.java
rnc: Ok

manipulation/src/main/java/org/jboss/gm/manipulation/actions/UploadTaskTransformerAction.java
rnc: From looking at https://github.com/bmuschko/gradle-nexus-plugin (now archived) I think the different Upload if check is fine.

While I might be ok with removing some of the checks I would rather this was done in a separate PR.

Given I don't think its possible to hit this `disabled` check in Gradle 7 can we remove it please.

cli/build.gradle.kts
rnc: Ok that makes sense. Can we add a sentence or so as comments describing why this extra code was needed please

manipulation/build.gradle.kts
rnc: Is it possible to put all the extra dependency instructions in the top level build file in one place?

rnc: Actually, I guess if we already have stuff in those files then that comment can be ignored :-)

rnc: Perhaps similar to my prior comment ; a single line explaining why we needed this change?

dwalluck: I wasn't sure how to reference the tasks in the main file, either, when I tried.

In that case, I will have to add a comment to three or four different files.

rnc: +1, a small comment will make future maintenance easier to understand why things were added IMHO

dwalluck: OK, I also did some slight rearranging so they're all located at the very bottom of their respective files now, with the same comment before the section.

build.gradle.kts
```diff
@@ -13,8 +13,14 @@ plugins {
     signing
     `maven-publish`
     idea
-    id("com.diffplug.gradle.spotless") version "3.25.0"
-    id("com.gradle.plugin-publish") version "0.12.0"
+
+    if (org.gradle.util.GradleVersion.current() < org.gradle.util.GradleVersion.version("5.4")) {
```
dwalluck, line 17: @rnc Why is the fully qualified name required here? If I try `GradleVersion.current()` it says not found. It would make these lines a bit easier to read.

rnc: Hmm, not sure I noticed this before and sort of ignored it :-)

rnc (approved):

build.gradle.kts
dwalluck: And the IDE is suggesting to remove it, but if I do, it breaks. I see it working in other examples, so I don't understand.

rnc: Frankly, not worth bothering about in _this_ PR ; can always investigate another time IMHO. I'm merging this as all checks have now passed.

PR #323, opened 2021-08-08
Preserve input argument order

dwalluck:


cli/src/main/java/org/jboss/gm/cli/Main.java
```diff
@@ -201,8 +203,9 @@ private void executeGradle() throws ManipulationException {
             Set<String> jvmArgs = jvmPropertyParams.entrySet().stream()
                     .map(entry -> "-D" + entry.getKey() + '=' + entry.getValue())
                     .collect(Collectors.toCollection(LinkedHashSet::new));
-            jvmArgs.addAll(ManagementFactory.getRuntimeMXBean().getInputArguments().stream()
-                    .filter(s -> s.startsWith("-Xdebug") || s.startsWith("-Xrunjdwp")).collect(Collectors.toSet()));
+            Collection<String> inputArguments = new ArrayList<>(ManagementFactory.getRuntimeMXBean().getInputArguments());
+            inputArguments.removeIf(s -> !s.startsWith("-Xdebug") || !s.startsWith("-Xrunjdwp"));
+            jvmArgs.addAll(inputArguments);
```
rnc, line 208: Can you elaborate why this is needed? Its only adding two arguments, the order of which doesn't matter e.g.
```
-Xdebug -Xrunjdwp:transport=dt_socket,server=y,suspend=y,address=5005
```

dwalluck: On the previous line, I went to the trouble of using a `LinkedHashSet` to get predictable order. It seemed to be more consistent and in line with that change to keep the order here as well (since the original data structure is a actually a `List`).

And normally with command-line arguments they can be repeated too, but we can't "fix" that unless we change from using `Set` to using `List` for `jvmArgs`.

rnc: Its not possible to duplicate the `-Xrunjdwp` argument - the JVM doesn't allow it. And the repeating the `Xdebug` argument, using a set means its normalised to one. The order of it doesn't matter.
Therefore I think this PR should be rejected.

dwalluck: In this narrow case, if truly  ` -Xrunjdwp` cannot be repeated and `-Xdebug` doesn't take any additional parameters the way ` -Xrunjdwp` can, then the only issue we really hit is the order and potentially experiencing a different order than what was originally passed, which to me is already bad enough.

As a general rule, repeating any arguments (for example, if `-Xdebug` could take different parameters) with a `Set` can lead to incorrect behavior. Using `HashSet` over `LinkedHashSet` makes this problem even worse. The original data structure is `ArrayList` which is ordered and allows repetition, so the closest we have is `LinkedHashSet` which is ordered.

Most importantly, consider the case of more than one `-Xdebug` (or whatever option). With command-line arguments, the last option should take precedence. Then, identical `-Xdebug` arguments would be consolidated, which already creates a problem. Using `HashSet` introduces an additional problem where even a different `-Xdebug` argument won't have the proper precedence.

Even `LinkedHashSet` cannot handle something like `-Xdebug1 -Xdebug2 -Xdebug1` correctly because 2 would have precedence over 1 when it should not, but at least it would handle the case of `-Xdebug1 -Xdebug2` where 2 should have precedence over 1 and does in this case.

rnc: You can only have `-Xdebug` not `-Xdebug1` or `-Xdebug=foobar`. So I don't think that applies in this case. The order is irrelevant as its only two parameters. There is no precedence involved. I'm sorry but I'm don't believe this PR is appropriate.

dwalluck: OK, we got off track. The real issue is not this part, but rather wanting to change from `LinkedHashSet` in my previous PR to `List` which I think would have been a better choice all along. I have updated the PR and now there are even less changes.

cli/src/main/java/org/jboss/gm/cli/Main.java
```diff
@@ -198,11 +197,11 @@ private void executeGradle() throws ManipulationException {
...
 
             BuildLauncher build = connection.newBuild();
-            Set<String> jvmArgs = jvmPropertyParams.entrySet().stream()
+            Collection<String> jvmArgs = jvmPropertyParams.entrySet().stream()
                     .map(entry -> "-D" + entry.getKey() + '=' + entry.getValue())
-                    .collect(Collectors.toCollection(LinkedHashSet::new));
+                    .collect(Collectors.toList());
             jvmArgs.addAll(ManagementFactory.getRuntimeMXBean().getInputArguments().stream()
-                    .filter(s -> s.startsWith("-Xdebug") || s.startsWith("-Xrunjdwp")).collect(Collectors.toSet()));
+                    .filter(s -> s.startsWith("-Xdebug") || s.startsWith("-Xrunjdwp")).collect(Collectors.toList()));
```
rnc, line 204: I'd rather this stayed as `Collectors.toSet()`

PR #326, opened 2021-09-07
Add support for Gradle 6.9.1 and Gradle 7.2

dwalluck:


cli/src/test/java/org/jboss/gm/cli/DifferentJVMTest.java
```diff
@@ -291,19 +294,20 @@ public void runWithJDK8Fails() throws Exception {
...
 
-        File gmeGradle = new File(projectRoot.getParentFile().getAbsolutePath(), "gme.gradle");
+        File gmeGradle = new File(projectRoot.getParentFile().getAbsolutePath(), AlignmentTask.GME);
         assertFalse(gmeGradle.exists());
         // XXX: Caused by: java.io.StreamCorruptedException: invalid type code: 00
         assumeTrue(v.compareTo(GradleVersion.version("5.3.1")) != 0);
         assertTrue(systemErrRule.getLog().contains("Task 'FOOgenerateAlignmentMetadataBAR' not found in root project"));
         // XXX: Gradle 4.x doesn't trigger the same Exception
-        assumeTrue(v.compareTo(GradleVersion.version("5.0")) >= 0);
+        assumeTrue(v.compareTo(GradleVersion.version("5.0")) >= 0 && v.compareTo(GradleVersion.version("7.2")) < 0);
```
rnc, line 310: Why are we ignoring Gradle versions > 7.1 ?

dwalluck: It doesn't pass this check in 7.2 (same as in 4.10) because the same Exception is not thrown. I suppose I should have changed the comment as well.

cli/src/test/java/org/jboss/gm/cli/DifferentJVMTest.java
```diff
@@ -291,19 +294,20 @@ public void runWithJDK8Fails() throws Exception {
         try {
             m.run(args);
         } catch (Exception e) {
-            if (v.compareTo(GradleVersion.version("5.0")) >= 0) {
-                // XXX: Ignored for 4.10 as Exception differs between Gradle versions and System.err is checked later anyway
+            if (v.compareTo(GradleVersion.version("5.0")) >= 0 && v.compareTo(GradleVersion.version("7.2")) < 0) {
```
dwalluck, line 297: More specifically, if it doesn't throw the same exception here, then the check later on also fails. I think that this code is relying or Gradle internals to be the same from release to release. It's obviously throwing an exception, or the test would fail in general, but it's not always the same code path.

cli/src/test/java/org/jboss/gm/cli/DifferentJVMTest.java
rnc: Does this mean it wouldn't 'pass' the `assertTrue(systemOutRule.getLog().contains("Java home overridden to: " + JDK8_DIR));` ? 
Frankly the Gradle constant change of their exception handling is a pain :-(

rnc:
Testing using e.g. version 7.0 it takes a slightly different path and throws `} catch (GradleConnectionException e) {` which is caught in Main.java around line 228. There is no useful cause there as its a continuous nested `StreamCorruptedException`.  Its possible 7.2 has finally fixed this issue.

There is the question of whether we should unwrap the LocationAwareException with e.g. 
```
if ("org.gradle.internal.exceptions.LocationAwareException".equals(exception.getClass().getName()) ||
                exception.toString().startsWith("org.gradle.internal.exceptions.LocationAwareException")
                ) {
```

Finally the final `assertTrue(systemOutRule.getLog().contains("Java home overridden to: " + JDK8_DIR));` **does** appear in the log for me although `Build exception but unable to transfer message due to mix of JDK versions. Examine log for problems` does not.

cli/src/test/java/org/jboss/gm/cli/DifferentJVMTest.java
rnc: Now for 7 the actual exception is `org.gradle.internal.exceptions.LocationAwareException: Task 'FOOgenerateAlignmentMetadataBAR' not found in root project 'junit6682355779458076435'.` although whether it should be unwrapped as per my other comment is another question

cli/src/test/java/org/jboss/gm/cli/DifferentJVMTest.java
rnc: I think this then stops the final assert working - see my other comment.

PR #329, opened 2021-10-01
NCL-6130: Add support for PREPARSE stage which is equivalent to FIRST

dwalluck:


build.gradle.kts
```diff
@@ -193,7 +193,7 @@ subprojects {
     extra["logbackVersion"] = "1.2.3"
     extra["mavenVersion"] = "3.5.0"
     extra["ownerVersion"] = "1.0.12"
-    extra["pmeVersion"] = "4.5"
+    extra["pmeVersion"] = "4.6-SNAPSHOT"
```
dwalluck, line 196: Needs final version?

common/src/main/java/org/jboss/gm/common/groovy/BaseScript.java
```diff
@@ -49,9 +49,10 @@
      */
     @Override
     public Project getProject() {
-        if (stage == InvocationStage.FIRST) {
-            logger.error("getProject unsupported for InvocationStage FIRST");
-            throw new ManipulationUncheckedException("getProject is not supported for Groovy in initial stage.");
+        if (stage == InvocationStage.PREPARSE || stage == InvocationStage.FIRST) {
+            logger.error("getProject unsupported for InvocationStage {}", stage.name());
+            throw new ManipulationUncheckedException(
```
dwalluck, line 54: I slightly changed the exception message to make this one and the one below consistent.

common/src/main/java/org/jboss/gm/common/groovy/BaseScript.java
```diff
@@ -134,7 +134,7 @@ public void setValues(Logger logger, InvocationStage stage, File rootDir, Proper
 
         File repoCache;
         try {
-            if (stage == InvocationStage.FIRST || stage == InvocationStage.BOTH) {
+            if (rootProject == null) {
```
dwalluck, line 137: At first I was going to do something like `if (stage != InvocationStage.LAST)`, but I think this check is even more correct than checking the stage.

build.gradle.kts
rnc: Yes it will eventually. I'm ok with snapshot during development.

common/src/main/java/org/jboss/gm/common/utils/GroovyUtils.java
```diff
@@ -90,7 +90,7 @@ public static void runCustomGroovyScript(Logger logger, InvocationStage targetSt
                 throw new ManipulationException("Mandatory annotation '@InvocationPoint(invocationPoint = ' not declared");
```
rnc, line 90: Does this suffer from the same problem where ALL is being passed in not the actual stage point?

common/src/main/java/org/jboss/gm/common/groovy/BaseScript.java
rnc: Ok

common/src/main/java/org/jboss/gm/common/utils/GroovyUtils.java
dwalluck: Yes. I will fix it.

dwalluck: I just changed it to pass in `targetStage` (stage it runs at) instead of `stage` (annotated).

PR #330, opened 2021-10-01
NCL-6130: Update invocation stage documentation

dwalluck:


guide/groovy.md
```diff
@@ -35,8 +35,8 @@ import org.jboss.gm.common.groovy.BaseScript
 
 ```
 
-where InvocationStage may be `FIRST`, `LAST` or `BOTH`. This denotes whether the script is ran
-before all other manipulators, after or both. The script therefore encodes how and when it is run.
+where `InvocationStage` may be `FIRST` (`PREPARSE`), `LAST` or `ALL`. This denotes whether the script is run
+before all other manipulators, after all other manipulators, or during all possible stages, respectively. The script therefore encodes how and when it is run.
```
rnc, line 39: I think we need to make it clear, that unlike PME, PREPARSE is equivalent to FIRST for GME.

dwalluck: I was thinking of writing "(`PREPARSE`, which is equivalent to `FIRST`)", but I thought it was clear by the parentheses that they are aliases.

rnc: Oh I've just thought of something - should the information about preparse be noted abuut when it is available?

dwalluck: Are the docs for all versions or just for the current version? I guess we do not have tagged docs specific to certain versions, So, I guess we need to still mention `BOTH` as well.

dwalluck: New version which uses similar table layout as in PME.

rnc (approved):

PR #331, opened 2021-10-05
Add missing javadoc comments

dwalluck:


dwalluck:
Java 17 seems to report by default any missing javadoc comments on `public` or `protected` entities, so I added it.

PR #342, opened 2021-10-28
NCL-6118: Add explicit support for supported version options

dwalluck:


common/src/main/java/org/jboss/gm/common/Configuration.java
```diff
@@ -73,6 +73,19 @@ default Properties getProperties() {
...
+
+    @Key("versionOverride")
+    String versionOverride();
+
+    @Key("versionSuffix")
+    String versionSuffix();
+
+    @Key("versionSuffixAlternatives")
+    String versionSuffixAlternatives();
+
```
rnc, line 88: I am in two minds about these. Technically they override what can be picked up already within the PME code. However saying that, it does make it explicit which is easier to follow for maintenance. Perhaps adding a comment describing that these are merely here to make it explicit?

rnc (commented):
Looks good, one small comment

dwalluck:
> Looks good, one small comment

Your thoughts on the wording in 73863ad? That's the only change. For some reason I had to force push (I maybe rebased).

rnc (approved):

PR #343, opened 2021-11-04
NCL-5172: Add documentation about unresolved dependencies

dwalluck:


index.md
```diff
@@ -118,7 +118,23 @@ When running the alignment plugin (e.g. via the CLI), it is possible to configur
 
 #### Unresolved Dependencies
 
-If the tool is not able to resolve certain dependencies then it may fail during the alignment phase. Set `ignoreUnresolvableDependencies` to true to ignore those (default: false).
+By default, the alignment phase is set to fail when it encounters any unresolvable dependencies. For example, if the
+dependency `com.redhat:unresolved-dependency` in configuration `default` is unable to be resolved, then the alignment
+phase will fail with an error message indicating which dependencies were unresolvable.
```
rnc, line 123: I wonder if we she clarify why dependencies might be unresolvable (and therefore why we use a LenientConfiguration internally).

dwalluck: The previous documentation simply said _may fail_ during the alignment.

Besides a bad dependency with an invalid groupId, artifactId, or version, what else might be the cause?

The other issue I see is suppose that an artifact is temporarily unresolvable (say, due to a network error). Is this possible? In such a case, the alignment would actually be missing something.

So, what are the valid reasons?

rnc: I actually can't remember what causes unresolvable deps within Gradle ; that was the point of the ticket to give this explanation to our users.

dwalluck: Maybe that example is not clear enough, but it is supposed to show an empty version. This is the specific case that I am sure about. I just saw it on a support ticket, and it is the only case appearing in one of our tests as well.

rnc: Agreed. Resolving this thread.

rnc (approved):

PR #345, opened 2021-11-26
NCL-6710: Add support for removing plugins

rnc:


cli/src/main/java/org/jboss/gm/cli/Main.java
```diff
@@ -286,10 +287,12 @@ public Void call() throws ManipulationException {
             if (logger.isDebugEnabled()) {
                 logger.debug("Configuration reloaded with {}", configuration.dumpCurrentConfig());
             }
-
-            GroovyUtils.runCustomGroovyScript(logger, InvocationStage.FIRST, target, configuration, null, null);
         }
 
+        GroovyUtils.runCustomGroovyScript(logger, InvocationStage.FIRST, target, configuration, null, null);
```
dwalluck, line 292: Was this a bug inside the `if` or does it need to have `groovyScripts` defined anyway?

cli/src/test/java/org/jboss/gm/cli/MainTest.java
```diff
@@ -70,6 +71,7 @@ public void reset() {
         // Reset the daemon between tests : https://discuss.gradle.org/t/stopping-gradle-daemon-via-tooling-api/16004/2
         // Under 4.10 the daemon appears to cache Config values which corrupt the tests.
         ConnectorServices.reset();
+        ConfigCache.getOrCreate(Configuration.class).reload();
```
dwalluck, line 74: So both `reset` and `reload` are required? I can only go by the original comment that "the daemon appears to cache Config values which corrupt the tests".

common/src/main/java/org/jboss/gm/common/utils/PluginUtils.java
```diff
@@ -0,0 +1,111 @@
...
+
+                    try {
+                        org.apache.commons.io.FileUtils.readLines(buildFile, Charset.defaultCharset());
+                        String target = org.apache.commons.io.FileUtils.readFileToString(buildFile, Charset.defaultCharset());
+                        String configTask = SUPPORTED_PLUGINS.get(plugin);
+
+                        logger.debug("Looking to remove {} with configuration block of {}", plugin, configTask);
+                        if (target.contains(plugin)) {
+
+                            List<String> lines = new ArrayList<>(Arrays.asList(target.split("\\R")));
```
dwalluck, line 62: In this different from using `org.apache.commons.io.FileUtils.readLines` directly instead of using `org.apache.commons.io.FileUtils.readFileToString`?

common/src/main/java/org/jboss/gm/common/utils/PluginUtils.java
```diff
@@ -0,0 +1,111 @@
...
+                            lines.removeIf(i -> i.contains(plugin));
+                            target = String.join("\n", lines);
```
dwalluck, line 64: OK, so technically we could lose newlines depending on OS. Not so much to worry about here unless we already had it in say PME `FileIO` already. Or, at least I remember adding something for JSON files there at one point.

common/src/test/java/org/jboss/gm/common/utils/PluginUtilsTest.java
```diff
@@ -0,0 +1,138 @@
...
+    @Rule
+    public final TestRule restoreSystemProperties = new RestoreSystemProperties();
+
+    private final Logger logger = LoggerFactory.getLogger(getClass());
+
+    @Before
+    public void reset() {
+        // Reset the daemon between tests : https://discuss.gradle.org/t/stopping-gradle-daemon-via-tooling-api/16004/2
+        // Under 4.10 the daemon appears to cache Config values which corrupt the tests.
+        ConnectorServices.reset();
```
dwalluck, line 47: For example, here is a `reset` without the `reload` but I did not see the original cause to be able to answer one way or another.

common/src/main/java/org/jboss/gm/common/utils/PluginUtils.java
rnc: Yes, I needed it both as individual lines and as a entire string. I'll clean it up a bit

common/src/main/java/org/jboss/gm/common/utils/PluginUtils.java
rnc: I've pushed https://github.com/release-engineering/pom-manipulation-ext/pull/897 in order to use the original newline.

cli/src/test/java/org/jboss/gm/cli/MainTest.java
rnc: I believe we both investigated this before. I'm now seeing other spurious errors and forcing a reload made them go away.

common/src/test/java/org/jboss/gm/common/utils/PluginUtilsTest.java
rnc: I only appeared to need the reset here

dwalluck: I believe it has something to do with the CLI vs running from a test. Nevermind then.

cli/src/test/java/org/jboss/gm/cli/MainTest.java
rnc: Experimenting with a `ConfigCache.clear` instead which might be more consistent.

rnc: Updated.

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
```diff
@@ -366,6 +367,17 @@ private void align(Configuration configuration, ManipulationCache cache, Manipul
         processAlignmentReport(rootProject, configuration, cache, nonAligned);
     }
 
+    private void processPropertiesForBuildCache(File rootProject) throws IOException {
+        File properties = new File(rootProject, "gradle.properties");
+        if (properties.exists()) {
+            List<String> lines = org.apache.commons.io.FileUtils.readLines(properties, Charset.defaultCharset());
+            if (lines.stream().anyMatch(s -> s.contains("org.gradle.caching"))) {
+                lines.removeIf(i -> i.contains("org.gradle.caching"));
+                FileUtils.writeLines(properties, lines);
```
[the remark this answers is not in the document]
rnc: I hadn't realise `removeIf` returns a boolean so yes I can simplify this.

analyzer/src/functTest/java/org/jboss/gm/analyzer/alignment/SimpleProjectFunctionalTest.java
```diff
@@ -111,6 +111,11 @@ public void ensureAlignmentFileCreated() throws IOException, URISyntaxException,
         assertThat(extraGradleFile).exists();
         assertThat(FileUtils.readLines(extraGradleFile, Charset.defaultCharset()))
                 .filteredOn(l -> l.trim().equals(AlignmentTask.APPLY_GME_REPOS)).hasSize(1);
+
+        // Verify the org.gradle.caching has been removed.
+        File properties = new File(projectRoot, "gradle.properties");
+        assertThat(FileUtils.readLines(properties, Charset.defaultCharset()))
+                .filteredOn(l -> l.contains("org.gradle.caching")).hasSize(0);
```
dwalluck, line 118: AssertJ does have a quicker/more compact way of handling files, so that you don't have to "manually" read the lines first:

```java
assertThat(linesOf(extraGradleFile)).containsOnlyOnce(AlignmentTask.APPLY_GME_REPOS);
```

```java 
assertThat(linesOf(properties)).doesNotContain("org.gradle.caching");
```

rnc: Nice! Thanks, have updated the new line.

PR #346, opened 2021-11-29
NCL-6710: Add support for removing plugins

rnc:


guide/plugins.md
```diff
@@ -0,0 +1,36 @@
...
+title: Plugins
+---
+
+* Contents
+{:toc}
+
+### Plugin Removal
+
+GME supports removal of plugins prior to running the alignment phase. Note that plugin removal is only supported through
+use of the CLI. If the property `-DpluginRemoval=<value>,....` is set, it will activate removal for the specified plugin.
```
dwalluck, line 11: Maybe instead of `value`, we could write `plugin` or `id`? You have "Plugin" as the column name. What is the official name for this, the plugin id? The point being that they are unique identifiers that come from the plugin itself.

I am suggesting adding a sentence that would make it clear what the value of the `pluginRemoval` property is. it's obviously meant to come from (one or more) values in the first table column but that isn't explicitly mentioned when the option is introduced.

rnc: Agreed. Improved docs

dwalluck (approved):

PR #365, opened 2022-03-01
NCLSUP-588 Refactor dependencyOverride handling to correct module han…

rnc:
…dling and simplify codebase


analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
```diff
@@ -273,15 +275,26 @@ private void align(Configuration configuration, ManipulationCache cache, Manipul
...
+                .sorted(comparingInt(AlignmentService.Manipulator::order))
+                .collect(Collectors.toList());
 
         final Response alignmentResponse = alignmentService.align(
                 new AlignmentService.Request(cache.getProjectVersionRefs(configuration.versionSuffixSnapshot()),
                         allDeps));
 
+        // Apply the current manipulators (DependencyOverride and UpdateProjectVersion)
+        // While they do support order, it's not hugely important given we only have two
+        // currently.
```
dwalluck, line 293: You sort the manipulators on line 281, but this comment seems to imply they are unsorted.

dwalluck (changes requested):

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
rnc: As those two manipulators don't affect each other the order currently isn't important. I could have stripped the ordering out but left it in for any future expansion.

dwalluck (approved):

PR #378, opened 2022-06-28
NCL-7148: Fix disabling version modification

dwalluck:


manipulation/src/main/java/org/jboss/gm/manipulation/ManipulationPlugin.java
```diff
@@ -91,24 +91,29 @@ public void apply(Project project) {
         final ManipulationModel correspondingModule = ManipulationIO.readManipulationModel(project.getRootDir())
                 .findCorrespondingChild(project);
 
-        // we need to change the project version early so various tasks that ready early and create other vars based on it
-        // (like the zip tasks) can use the correct version
-        logger.info("Updating project ({}) version {} to {}", project.getProjectDir(), project.getVersion(),
-                correspondingModule.getVersion());
-        project.setVersion(correspondingModule.getVersion());
+        if (!project.getVersion().equals(correspondingModule.getVersion())) {
```
dwalluck, line 94: @rnc Is this check OK, or is it possible to pass this check and still fail `if (!correspondingModule.getVersion().equals(project.getVersion()))`? I needed some way to check for `versionModification` from the manipulation plugin.

rnc (approved):

PR #385, opened 2022-07-10
NCL-7182: Inherit environment variables

dwalluck:


dwalluck:
All tests pass, so if the issue in 6e2c842e1bdccdb7b7c12453d7f51d29eba7d6b8 (https://github.com/gradle/gradle/issues/3117) is being tested, that is fixed as well.

cli/src/main/java/org/jboss/gm/cli/Main.java
```diff
@@ -11,6 +11,7 @@
 import java.nio.file.StandardOpenOption;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
```
rnc, line 14: I think Pair/Stream imports are no longer used

cli/src/main/java/org/jboss/gm/cli/Main.java
```diff
@@ -227,6 +220,8 @@ private void executeGradle(OutputStream out, boolean quiet, List<String> current
             logger.info("Executing CLI {} on Gradle project {} with JVM args '{}' and arguments '{}'",
                     ManifestUtils.getManifestInformation(Main.class), target, jvmArgs, currentGradleArgs);
 
+            logger.debug("Environment variables: {}", envVars);
```
rnc, line 223: This ends up (at least on my system) dumping a **large** amount of extra output - around 132231 characters. While I don't think this is suitable for trace level output would a compromise be to add
```
envVars.remove("LS_COLORS");
```
which reduces it on my shell down to 4186 characters?

It could be argued that PROPT/RPROMPT/PS1 could also be removed which has the side benefit of removing any strange control characters in case they cause an issue (which reduces down to 3946 on my system) : 
```
            envVars.keySet().removeIf(k -> k.matches( "^(LS_COLORS|RPROMPT|PROMPT|PS1)$" ));
```

cli/src/main/java/org/jboss/gm/cli/Main.java
dwalluck: Done in [9458940](https://github.com/project-ncl/gradle-manipulator/pull/385/commits/945894036cc17e2f333fcb61f5e574ad27792d83).

cli/src/main/java/org/jboss/gm/cli/Main.java
dwalluck: The logger is needed for the test. Otherwise, I would just remove it. I don't think there's a way to probe the environment programmatically. I don't even see a `BuildLauncher::getEnvironmentVariables`.

cli/src/main/java/org/jboss/gm/cli/Main.java
```diff
@@ -227,6 +218,12 @@ private void executeGradle(OutputStream out, boolean quiet, List<String> current
             logger.info("Executing CLI {} on Gradle project {} with JVM args '{}' and arguments '{}'",
                     ManifestUtils.getManifestInformation(Main.class), target, jvmArgs, currentGradleArgs);
 
+            if (logger.isDebugEnabled()) {
+                Map<String, String> envVarsToPrint = new HashMap<>(envVars);
+                envVarsToPrint.keySet().removeIf(k -> k.matches("^(LS_COLORS|RPROMPT|PROMPT|PS1)$"));
+                logger.debug("Environment variables: {}", envVarsToPrint);
+            }
```
rnc, line 225: This isn't quite what I meant - I meant remove those aforementioned variables completely and avoid passing them through to Gradle - which then also reduces the debug output.

dwalluck: Sorry, I thought you were referring to the log output only in your comment. Why do we want to filter any of the variables passed in, which depending on the environment may not even be set in the first place?

One thing I am trying to do with this PR is make launching Gradle from `Main` the same as if it was launched from the command-line. From the command line, it can apparently see all environment variables.

I originally thought it had something to do with forking/daemon, but that didn't seem to change the behavior.

dwalluck: That's sort of the inverse of the problem in original fix which only passed in a few particular environment variables and dropped all others. This means that our process launched from `Main` cannot see what the build itself will see even if we run it on the same system one right after the other.

dwalluck: According to the Gradle code, it does not print the values as they main contain sensitive information. This should also solve the too much output issue.

rnc: I'll try this tomorrow but at first glance if its the same as Gradle internals then it should be ok. My only concern is older 4.10 runs and if our prompts contain non-ascii symbols - I'll try it tomorrow.

dwalluck: I think I originally had tried this in the test. I will update the test. It's just make any variable contain non-ASCII symbols, right?

rnc: I've tried it with an older build (fge/btf.git) and using 4.10.2 e.g.

```
java -jar ~/Work/Gradle/gradle-manipulator/cli/build/libs/cli-3.7-SNAPSHOT.jar -d  --init-script=~/Work/Gradle/gradle-manipulator/analyzer/build/resources/main/analyzer-init.gradle -DdependencySource=NONE generateAlignmentMetadata -l=~/Software/gradle-4.10.2
```
With a prompt containing symbols e.g. `%F{red}🎩%` running the above crashes out before even processing the build. Removing the prompt/etc variables it at least processes the build, but then the daemon crashes out at the end (which is what it did before).

cli/src/main/java/org/jboss/gm/cli/Main.java
dwalluck: Resolved by using `keySet()`.

cli/src/main/java/org/jboss/gm/cli/Main.java
dwalluck: I had to rearrange some code to reuse the code that was already there for the version check. I updated the log message to also include the URL to the upstream issue, if that's OK.

Sorting is done exactly as suggested here.

rnc (approved):

PR #386, opened 2022-07-14
NCL-7185: Ignore overrideTransitive if dependencySource=NONE

dwalluck:
If GME detects the shadow plugin, it currently requires the user to explicitly set `overrideTransitive`. However, if `dependencySource=NONE`, then dependencies won't get aligned anyway, so I don't think the build should fail in that case.


analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
```diff
@@ -592,11 +593,13 @@ private Map<RelaxedProjectVersionRef, ProjectVersionRef> getDependencies(Project
                 }
                 Set<ResolvedDependency> target;
                 if (internalConfig.overrideTransitive() == null || !internalConfig.overrideTransitive()) {
```
dwalluck, line 595: This is the original code, but it looks like this outer `if` was never necessary?

PR #389, opened 2022-07-19
NCLSUP-526 NCLSUP-538 Inline Dokka plugin handling and make version d…

rnc:
…ependent

### DIFF
--- a/analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
+++ b/analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
@@ -20,6 +20,8 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -43,6 +45,7 @@ import org.commonjava.maven.ext.common.util.JSONUtils;
 import org.commonjava.maven.ext.core.groovy.InvocationStage;
 import org.commonjava.maven.ext.core.state.DependencyState;
 import org.gradle.api.DefaultTask;
+import org.gradle.api.Plugin;
 import org.gradle.api.Project;
 import org.gradle.api.artifacts.Dependency;
 import org.gradle.api.artifacts.DependencySet;
@@ -53,7 +56,9 @@ import org.gradle.api.artifacts.UnresolvedDependency;
 import org.gradle.api.artifacts.repositories.ArtifactRepository;
 import org.gradle.api.internal.artifacts.configurations.ConflictResolution;
 import org.gradle.api.internal.artifacts.ivyservice.resolutionstrategy.DefaultResolutionStrategy;
+import org.gradle.api.internal.plugins.DefaultPluginManager;
 import org.gradle.api.logging.Logger;
+import org.gradle.api.plugins.AppliedPlugin;
 import org.gradle.api.publish.PublishingExtension;
 import org.gradle.api.publish.maven.MavenPublication;
 import org.gradle.api.tasks.TaskAction;
@@ -68,6 +73,7 @@ import org.jboss.gm.common.io.ManipulationIO;
 import org.jboss.gm.common.logging.GMLogger;
 import org.jboss.gm.common.model.ManipulationModel;
 import org.jboss.gm.common.utils.GroovyUtils;
+import org.jboss.gm.common.utils.PluginUtils.DokkaVersion;
 import org.jboss.gm.common.utils.ProjectUtils;
 import org.jboss.gm.common.versioning.DynamicVersionParser;
 import org.jboss.gm.common.versioning.ProjectVersionFactory;
@@ -125,6 +131,8 @@ public class AlignmentTask extends DefaultTask {
      * The task name {@code generateAlignmentMetadata}.
      */
     public static final String NAME = "generateAlignmentMetadata";
+
+    private static final String DOKKA = "org.jetbrains.dokka";
 
     private static final ContextClassLoaderLocal<AtomicBoolean> configOutput = new ContextClassLoaderLocal<AtomicBoolean>() {
         @Override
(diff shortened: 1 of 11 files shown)


analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
```diff
@@ -246,6 +254,23 @@ public void perform() {
...
             }
 
+            AppliedPlugin ap = project.getPluginManager().findPlugin(DOKKA);
+            if (ap != null) {
+                logger.debug("Plugin {} has been applied to {}", ap.getId(), project.getName());
+
+                @SuppressWarnings("rawtypes")
+                Plugin p = ((DefaultPluginManager) project.getPluginManager()).getPluginContainer().findPlugin(DOKKA);
+                String path = p.getClass().getProtectionDomain().getCodeSource().getLocation().getPath();
+                Matcher m = Pattern.compile(".*-([\\d.]+[a-zA-Z]*)\\.jar").matcher(path);
```
dwalluck, line 264: While I think `[\\d.]+` matches, it would also match `"..."? Maybe something like `\d+(\.\d+)*`?

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
```diff
@@ -246,6 +254,23 @@ public void perform() {
...
+                if (m.find()) {
```
dwalluck, line 265: I think you want `matches()` instead of `find()` since `.jar` should be the very end of the string (equivalent to `\\.jar$`).

cli/src/test/resources/build.gradle
```diff
@@ -36,10 +36,10 @@ task api(type: Javadoc) {
     }
 }
 
-System.out.println("Environment: " + System.getenv())
 if ( System.getenv("JAVA_HOME") == null ) {
     throw new RuntimeException()
 }
+System.out.println("Environment: JAVA_HOME=" + System.getenv("JAVA_HOME"))
```
dwalluck, line 42: Not sure what this change is related to, but I see you're checking for it in the logs during the tests?

common/src/main/java/org/jboss/gm/common/utils/PluginUtils.java
```diff
@@ -45,6 +50,35 @@ public class PluginUtils {
...
+    public enum DokkaVersion {
+        NONE(""),
+        MINIMUM("0.9.18"),
+        TEN("0.10.0"),
+        POST_ONE("1.4.0");
+
+        final VersionSpec version;
+
+        DokkaVersion(String s) {
+            this.version = isEmpty(s) ? null : VersionUtils.createSingleVersion(s);
```
dwalluck, line 62: Does this make passing in `null` or `""` both equivalent to `NONE("")`? I just wonder why it's passed the empty string, but then `version` is set to `null`.

cli/src/test/resources/build.gradle
rnc: Your change to pass the full environment through broke this test on my machine which has a coloured prompt - I resolved it by limiting its printout to just JAVA_HOME instead of the entire env.

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/io/SettingsFileIO.java
```diff
@@ -78,6 +79,43 @@ public String writeProjectNameIfNeeded(File rootDir) throws IOException {
...
+                String settingsContents = FileUtils.readFileToString(settings, Charset.defaultCharset());
+                logger.info("Updating {} with Dokka resolutionStrategy information", settings);
+
+                if (settingsContents.contains("resolutionStrategy")) {
+                    // Existing strategy so inject another
+                    settingsContents = settingsContents.replaceFirst("(?s)resolutionStrategy(\\s|\\n|$)+\\{",
+                            "resolutionStrategy {\n eachPlugin { if (requested.id.id == \"org.jetbrains.dokka\") { useVersion(\"0.9.18\") } }\n");
+                } else if (settingsContents.contains("pluginManagement")) {
+                    // Existing pluginManagement but no strategy so inject that
+                    settingsContents = settingsContents.replaceFirst("(?s)pluginManagement(\\s|\\n|$)+\\{",
```
dwalluck, line 107: I think that `\\s` is already matching newlines (`\\n`). Also, then it should handle a case we probably won't hit in practice of a `\\r?\\n`.

rnc: This is the same pattern (I think) that I used successfully in PluginUtils.

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
rnc: I odn't think we could ever have a jar installed using `.....` though surely?

common/src/main/java/org/jboss/gm/common/utils/PluginUtils.java
rnc: I guess it could be `NONE(null)` ... and then `s == null ? null : ...` ?

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
dwalluck: Right, it probably won't fail in practice, just pointing out the `[.]+`.

cli/src/test/resources/build.gradle
dwalluck: I think you're referring to `DifferentJvmTest:224` failing?

We could possibly use logging at `Main.java:205-211` where it does the actual `JAVA_HOME` check.

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/io/SettingsFileIO.java
dwalluck: All I am saying is `\\s` already is "A whitespace character: `[ \t\n\x0B\f\r]`".

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
rnc: Yeah that sounds reasonable, will change, thanks.

cli/src/test/resources/build.gradle
rnc: The DifferentJVMTest was failing due to colours being printed by the `getenv` call. Simplifying this test resolves it.

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
rnc: Using https://regex101.com having played with e.g. 
```
gradle-enterprise-conventions-gradle-plugin-0.0.9.jar
bad-plugin-0.1..hello.jar
publish-plugin-1.1.0.jar
gradle-enterprise-conventions-gradle-plugin-0.0.9.jar
dependency-management-plugin-1.0.7.RELEASE.jar
propdeps-plugin-0.0.9.RELEASE2.jar
os-maven-plugin-1.4.0.Final.jar
weird-1.1.REL_11.jar
maven-settings-plugin-0.5.jar
```
a minor tweak produces `.*-([\d.]+[\w_]*)\.jar` which catches all of them and I think is reasonable.

analyzer/src/main/java/org/jboss/gm/analyzer/alignment/io/SettingsFileIO.java
rnc: Ah I see, thanks I hadn't realised whitespace included newline. I'll change both for consistency.

common/src/main/java/org/jboss/gm/common/utils/PluginUtils.java
rnc: Agreed, will change as per my above suggestion.

PR #391, opened 2022-07-20
Don't bother overwriting manifest version values if they haven't changed

dwalluck:
Although replacing values with identical values as we do now is harmless, it seemed important to not always do it, particularly when the version has not changed (`-DversionModification=false`).

I wonder about logging some of these things (non-changes) at info level as it may be a little too verbose.

I did change some of the log messages to make it more clear (to me) what is happening. I think I'd like to see the project name printed more often so that you don't have to find it in the output by scrolling back. Also, programmatically there's no easy way to get the project context from the log if it's not on the same line for testing purposes.

### DIFF
--- a/manipulation/src/functTest/resources/mongo-java-driver-no-overwrite/build.gradle
+++ b/manipulation/src/functTest/resources/mongo-java-driver-no-overwrite/build.gradle
@@ -1,0 +1,313 @@
+/*
+ * Copyright 2008-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+buildscript {
+    repositories {
+        mavenLocal()
+        mavenCentral()
+        maven { url "https://plugins.gradle.org/m2/" }
+    }
+    dependencies {
+        classpath 'com.netflix.nebula:gradle-extra-configurations-plugin:3.0.3'
+        classpath 'biz.aQute.bnd:biz.aQute.bnd.gradle:4.3.1'
+
+        // Scala plugins
+        classpath "com.adtran:scala-multiversion-plugin:1.0.35"
+        classpath "com.diffplug.spotless:spotless-plugin-gradle:3.27.1"
+    }
+}
+
+plugins {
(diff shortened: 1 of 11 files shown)


manipulation/src/functTest/resources/mongo-java-driver-no-overwrite/build.gradle
```diff
@@ -0,0 +1,313 @@
...
+def scalaProjects = subprojects.findAll { it.name.contains('scala') }
+def javaMainProjects = javaProjects.findAll { !['util'].contains(it.name) }
+def javaCodeCheckedProjects = javaMainProjects.findAll { !['driver-benchmarks'].contains(it.name) }
+
+configure(coreProjects) {
+    apply plugin: 'idea'
+
+//     evaluationDependsOn(':util')
+    group = 'org.mongodb'
+    version = '4.1.0.temporary-redhat-00001'
```
dwalluck, line 69: Changed from the original project so that the version did not change (4.1.0 -> 4.1.0.temporary-redhat-00001).

manipulation/src/functTest/resources/mongo-java-driver-no-overwrite/driver-core/build.gradle
```diff
@@ -0,0 +1,81 @@
...
+    clsName = 'MongoDriverVersion'
+    packageName = 'com.mongodb.internal.build'
+}
+
+jar {
+    manifest {
+        attributes("Implementation-Version": version,
+                   "Specification-Version": version)
+    }
+}
```
dwalluck, line 60: Added from the original project so that the original entries already exist at the time the manifest update action is run.

manipulation/src/functTest/resources/mongo-java-driver-no-overwrite/driver-core/build.gradle
```diff
@@ -0,0 +1,81 @@
...
+        'javax.security.auth.callback.*',
+        'org.ietf.jgss.*',
+        'io.netty.*;resolution:=optional',
+        'org.xerial.snappy.*;resolution:=optional',
+        'com.github.luben.zstd.*;resolution:=optional',
+        'org.slf4j.*;resolution:=optional',
+        'jnr.unixsocket.*;resolution:=optional',
+        'com.mongodb.crypt.capi.*;resolution:=optional'
+    ].join(',')
+}
```
dwalluck, line 81: This is from the original project. I saw this, and I am not sure this is having an effect at the time the manifest action is run (note the `afterEvaluate`).

manipulation/src/functTest/resources/mongo-java-driver-no-overwrite/manipulation.json
```diff
@@ -0,0 +1,21 @@
+{
+  "group" : "org.mongodb",
+  "name" : "gradle-manipulator",
+  "projectPathName" : "mongo-java-driver",
+  "version" : "4.1.0.temporary-redhat-00001",
+  "originalVersion" : "4.1.0.temporary-redhat-00001",
```
dwalluck, line 6: Changed versions from original project to match.

rnc (approved):

PR #395, opened 2022-07-21
Check for deploy URL and publish plugin sooner

dwalluck:
These checks can be done earlier to avoid running unnecessary code.


rnc:
Why not check in MavenPublishingRepositoryAction sooner as well ?

dwalluck:
> Why not check in MavenPublishingRepositoryAction sooner as well ?

Good point. Thanks.

We can also move the plugin check up. Now the two classes basically match, and we're not wasting time doing work when we're just going to return.

rnc (approved):

PR #406, opened 2022-08-01
Kotlin generation - avoid recursive configuration copying. Use double…

rnc:
… quotes. Improved buildscript injection


dwalluck (approved):
Looks good to me.

As a side note, I noticed a couple changes from my last PR that must not have been properly formatted and then were reformatted in this PR. Should we have the PR fail if that's the case?

PR #414, opened 2022-10-11
NCLSUP-766 Enhance plugin removal

rnc:


dwalluck:
@rnc I have run into at least one case of the regex removing too much and breaking a build. I wonder if we have the `clazz` for a task, why couldn't we use `project.tasks.withType(clazz) { enabled = false }`?

rnc:
> @rnc I have run into at least one case of the regex removing too much and breaking a build. I wonder if we have the `clazz` for a task, why couldn't we use `project.tasks.withType(clazz) { enabled = false }`?

Interesting point ; I haven't tried that. These changes (and some prior ones) were to try and avoid the problems you mentioned although there is always room for further improvement,

dwalluck (approved):

PR #417, opened 2022-10-13
Fix Dokka version detection

stuartwdouglas:
The noJdkLink attribute was added in 0.9.18, however this will always attempt to add this attribute even when dokka is an older version, as the existing code will never return NONE, only MINIMUM.

### DIFF
--- a/common/src/main/java/org/jboss/gm/common/utils/PluginUtils.java
+++ b/common/src/main/java/org/jboss/gm/common/utils/PluginUtils.java
@@ -71,12 +71,14 @@ public class PluginUtils {
             } catch (InvalidVersionSpecificationException e) {
                 throw new ManipulationException("Unable to parse version: {}", version);
             }
-            if (MINIMUM.version.compareTo(v) >= 0) {
-                return MINIMUM;
-            } else if (v.compareTo(POST_ONE.version) >= 0) {
+            if (v.compareTo(POST_ONE.version) >= 0) {
                 return POST_ONE;
+            } else if (v.compareTo(TEN.version) >= 0) {
+                return TEN;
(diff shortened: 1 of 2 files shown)


rnc:
Thanks for the PR @stuartwdouglas. Can you point me to the Gradle repository / version that is causing the issue?

In that code `NONE` was meant to represent no Dokka plugin. From what you've pointed out (and I can see in https://github.com/Kotlin/dokka/releases/tag/0.9.18 ) then I think the code in https://github.com/project-ncl/gradle-manipulator/blob/main/analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java#L488 needs to be changed to handle both 0.9.17 and 0.9.18 and the latter can have `noJdkLink`.

dwalluck:
> Thanks for the PR @stuartwdouglas. Can you point me to the Gradle repository / version that is causing the issue?

@rnc I believe it is https://github.com/square/wire/tree/3.7.1.

```
* What went wrong:
A problem occurred configuring project ':wire-compiler'.
> Could not set unknown property 'noJdkLink' for task ':wire-compiler:dokka' of type org.jetbrains.dokka.gradle.DokkaTask.
```

rnc:
Actually I'm now leaning towards thinking this is more subtle - it is meant to change the settings to force upgrade to 0.9.18 via
```
pluginManagement { resolutionStrategy { eachPlugin { if (requested.id.id == "org.jetbrains.dokka") { useVersion("0.9.18") } } } }
```
but that is obviously not working in wire.

rnc:
@stuartwdouglas Given my prior comment, do you have any objections/thoughts/etc if I close this? We can discuss this further offline in chat if you wish? (@dwalluck can point you to the channel)

PR #424, opened 2022-11-24
NCL-7432 Add OpenTelemetry instrumentation to GME

rnc:


build.gradle.kts
```diff
@@ -192,16 +192,17 @@ subprojects {
...
+    extra["ivyVersion"] = "2.5.0"
     extra["jacksonVersion"] = "2.11.2"
+    extra["jgitVersion"] = "6.3.0.202209071007-r"
     extra["junitVersion"] = "4.13.1"
     extra["logbackVersion"] = "1.2.9"
     extra["mavenVersion"] = "3.5.0"
+    extra["opentelemetryVersion"] = "1.2.0"
     extra["ownerVersion"] = "1.0.12"
-    extra["pmeVersion"] = "4.8"
+    extra["pmeVersion"] = "4.14-SNAPSHOT"
```
rnc, line 203: Need to release PME 4.14 first

common/src/main/java/org/jboss/gm/common/utils/OTELUtils.java
```diff
@@ -0,0 +1,51 @@
...
+            } finally {
+                FilteringCustomLogger.getContext().setLevel(originalLevel);
+            }
+        }
+    }
+
+    public void stopOTel() {
+        try {
+            if (originalLevel != LogLevel.DEBUG) {
+                FilteringCustomLogger.getContext().setLevel(LogLevel.DEBUG);
```
dwalluck, line 44: Should be `originalLevel`?

common/src/main/java/org/jboss/gm/common/utils/OTELUtils.java
```diff
@@ -0,0 +1,51 @@
...
+        if (endpoint != null) {
+            if (service == null) {
+                service = "gradle-manipulator";
+            }
+            logger.info("Enabling OpenTelemetry collection on {} with service name {} level {}", endpoint, service,
+                    originalLevel);
+            try {
+                if (originalLevel != LogLevel.DEBUG) {
+                    FilteringCustomLogger.getContext().setLevel(LogLevel.DEBUG);
+                }
```
dwalluck, line 32: The log message seems a little confusing to me as it prints `level originalLevel` in the log, but actually forces it to use `DEBUG`?

common/src/main/java/org/jboss/gm/common/utils/OTELUtils.java
dwalluck: Eh, no, you do set that in the finally. But, it should have already been set on start, line 36, so I wasn't sure.

common/src/main/java/org/jboss/gm/common/utils/OTELUtils.java
rnc: I was trying to always get the output from the opentelemetry-java-cli library as its useful to see the trace and span ids.

common/src/main/java/org/jboss/gm/common/utils/OTELUtils.java
rnc: In each start/stop there is a wrapper around the logging control.

common/src/main/java/org/jboss/gm/common/utils/OTELUtils.java
rnc: I've removed the confusing bit of the logging statement (left over debug that I missed)

dwalluck (approved):

PR #425, opened 2022-12-01
NCLSUP-793 Auto-disable signing plugin. Handle strict resolution stra…

rnc:
…tegy

### DIFF
--- a/common/src/testFixtures/java/org/jboss/gm/common/JVMTestSetup.java
+++ b/common/src/testFixtures/java/org/jboss/gm/common/JVMTestSetup.java
@@ -1,0 +1,72 @@
+package org.jboss.gm.common;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+import kong.unirest.Unirest;
+
+import org.apache.commons.lang.SystemUtils;
+import org.codehaus.plexus.archiver.UnArchiver;
+import org.codehaus.plexus.archiver.tar.TarGZipUnArchiver;
+import org.codehaus.plexus.archiver.zip.ZipUnArchiver;
+import org.codehaus.plexus.logging.LogEnabled;
+import org.codehaus.plexus.logging.console.ConsoleLoggerManager;
+import org.commonjava.maven.ext.common.ManipulationUncheckedException;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class JVMTestSetup {
+    public static final Path GRADLE_JDK_HOME = Paths.get(System.getProperty("user.home"), ".gradle", "jdks");
+
+    public static final String JDK8_BASEDIR = "jdk8u272-b10";
+
+    public static final Path JDK8_DIR = GRADLE_JDK_HOME.resolve(JDK8_BASEDIR);
+
+    public static final Path JDK8_BIN_DIR = JDK8_DIR.resolve("bin");
+
+    /**
+     * This method will, on Linux, download and cache if it doesn't exist a JDK8 installation from AdoptOpenJDK.
+     *
+     * This location ($HOME/.gradle/jdks) was chosen to match https://docs.gradle.org/current/userguide/toolchains.html.
+     * It utilises the same directory structure and location thereby matching potential prior downloads.
+     */
+    public static void setupJVM() throws IOException {
+        String filename = null;
+        UnArchiver ua = null;
+
+        if (SystemUtils.IS_OS_LINUX) {
+            filename = "OpenJDK8U-jdk_x64_linux_hotspot_8u272b10.tar.gz";
+            ua = new TarGZipUnArchiver();
+        } else if (SystemUtils.IS_OS_WINDOWS) {
+            filename = "OpenJDK8U-jdk_x64_windows_hotspot_8u272b10.zip";
+            ua = new ZipUnArchiver();
+        } else {
+            throw new ManipulationUncheckedException("Unknown OS");
+        }
+
+        if (!Files.exists(JDK8_BIN_DIR)) {
+            Files.createDirectories(JDK8_DIR);
+
+            Path destFile = GRADLE_JDK_HOME.resolve(filename);
+
+            destFile.toFile().deleteOnExit();
+
+            Unirest.get(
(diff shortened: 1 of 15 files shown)


common/src/testFixtures/java/org/jboss/gm/common/JVMTestSetup.java
```diff
@@ -0,0 +1,72 @@
...
+import org.codehaus.plexus.archiver.UnArchiver;
+import org.codehaus.plexus.archiver.tar.TarGZipUnArchiver;
+import org.codehaus.plexus.archiver.zip.ZipUnArchiver;
+import org.codehaus.plexus.logging.LogEnabled;
+import org.codehaus.plexus.logging.console.ConsoleLoggerManager;
+import org.commonjava.maven.ext.common.ManipulationUncheckedException;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class JVMTestSetup {
```
rnc, line 20: Made this a utility class so both CLI an Manipulation modules can use it.

manipulation/src/functTest/java/org/jboss/gm/manipulation/EHCacheProjectFunctionalTest.java
```diff
@@ -0,0 +1,86 @@
...
+                .setDirectory(simpleProjectRoot)
+                .setBranch("refs/tags/v3.10.2")
+                .setDepth(1)
+                .call()) {
+            System.out.println("Cloned ehcache3 to " + simpleProjectRoot);
+        }
+
+        // Now copy in overrides to enable this subset of unit test to work:
+        //   Modify Maven repository
+        //   Modify docs generation
```
rnc, line 60: The docs modification in clustered/ehcache-clustered is the only thing I haven't got working yet for ehcache.

manipulation/src/main/java/org/jboss/gm/manipulation/ManipulationPlugin.java
```diff
@@ -214,6 +233,30 @@ private void configurePublishingTask(Configuration config, Project project, Mani
...
+                        required = true;
+                    }
+                } catch (IllegalAccessException e) {
+                    throw new ManipulationUncheckedException(e);
+                }
+                if (required) {
+                    logger.warn("Signing was detected as enabled - disabling.");
+                    signingExtension.setRequired(false);
+                }
+            }
```
rnc, line 259: This block may make removing the signing plugin redundant - and we could perhaps consider disabling that by default then ?

manipulation/src/main/java/org/jboss/gm/manipulation/ManipulationPlugin.java
```diff
@@ -118,6 +124,19 @@ public void apply(Project project) {
...
+            if (c.isCanBeResolved() && c.getResolutionStrategy() instanceof DefaultResolutionStrategy) {
+                DefaultResolutionStrategy defaultResolutionStrategy = (DefaultResolutionStrategy) c.getResolutionStrategy();
+                if (defaultResolutionStrategy.getConflictResolution() == ConflictResolution.strict) {
+                    // failOnVersionConflict() sets this which causes our plugin to crash out. Reset to latest to make an attempt
+                    // at continuing. As Gradle creates 'decorated', we can't use reflection to change the value
+                    // back to the default. Therefore, use preferProjectModules as it's not eager-fail.
+                    logger.warn("Detected use of conflict resolution strategy strict in {}; resetting to preferProjectModules.",
+                            c.getName());
+                    defaultResolutionStrategy.preferProjectModules();
+                }
```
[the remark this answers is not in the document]
dwalluck: Sure. I think I see a difference of `c.isCanBeResolved()` added here vs. in `ProjectUtils`.

manipulation/src/main/java/org/jboss/gm/manipulation/ManipulationPlugin.java
dwalluck: I think it is better to do it in code if we can rather than relying on text modification.

manipulation/src/functTest/java/org/jboss/gm/manipulation/EHCacheProjectFunctionalTest.java
rnc: I've figured out now (and put details on NCLSUP-793) why docs is causing a problem. Due to them having a custom build plugin, it seems to conflict. I don't think this is something GME can handle but have suggested a solution for the product team using a groovy script.

manipulation/src/main/java/org/jboss/gm/manipulation/ManipulationPlugin.java
rnc: Ok, thanks, so this leads onto two further points:

* I found https://github.com/gradle/gradle/blob/master/subprojects/signing/src/main/java/org/gradle/plugins/signing/SigningExtension.java#L113 so I think this exists in >= Grade 4.10
  *  I think the signing type is always a Boolean but I'm not sure

* We currently have `pluginRemoval=ALL` as a shortcut. I think we should keep that but then that means we need a suitable extra shortcut e.g. 'recommended' (`REC`) or 'standard' (`STD`) etc. Can you suggest another good shortcut name and abbreviation to add ? Obviously this will also need a documentation update.

dwalluck: 1. Yes. https://github.com/gradle/gradle/blob/v4.10.3/subprojects/signing/src/main/java/org/gradle/plugins/signing/SigningExtension.java#L115
2. Maybe `DEFAULT`?

rnc: >     2. Maybe `DEFAULT`?

While I quite like default, one contextual problem is that it is not the default - plugins are not removed by default (whether they _should_ be is another discussion!) This is why I am leaning towards 'recommended'.

dwalluck: I meant default list of plugins. I mean, the option wouldn't appear unless you specified that option to begin with?

dwalluck: In any case, naming is not critical IMO.

dwalluck (approved):

PR #427, opened 2022-12-08
NCLSUP-800 Add BOM support

rnc:


analyzer/src/main/java/org/jboss/gm/analyzer/alignment/AlignmentTask.java
```diff
@@ -634,18 +637,20 @@ private Map<RelaxedProjectVersionRef, ProjectVersionRef> getDependencies(Project
                 // analyzer/src/functTest/java/org/jboss/gm/analyzer/alignment/DynamicWithLocksProjectFunctionalTest.java for
                 // an example) first verify if DefaultProjectDependencyConstraint occurs in the list of constraints.
                 LenientConfiguration lenient;
-                //
-                // Some kotlin projects (e.g. wire) have issues calling copyRecursive.
```
rnc, line 638: I've decided to remove this as I can't reproduce it on wire any more. Therefore if it reoccurs I'll deal with it then.

dwalluck: Makes sense.

PR #438, opened 2023-05-22
Improve handling of existing export-package

rnc:


grgrzybek (approved):

grgrzybek:
I built asm 9.5.0.redhat-00001 and see correct output using GME with the changes:
```
[ManifestUpdateAction.java:154] For project asm, task jar, not updating Export-Package since version (9.5.0.redhat-00001) has not changed
```

rnc:
For future reference - all CI passed bar codecov which seems to be playing up even though I have added _more_ tests.

PR #440, opened 2023-05-25
Fix coverage running manipulation tests via GradleRunner enabling debug.

rnc:


manipulation/src/functTest/java/org/jboss/gm/manipulation/TestUtils.java
```diff
@@ -115,12 +115,19 @@ static GradleRunner createGradleRunner() {
             jvmArgs.add("-D" + a + '=' + System.getProperty(a));
         }
 
-        jvmArgs.add("-DgmeFunctionalTest=true"); // Used to indicate for the plugin to reinitialise the configuration.
+        // Used to indicate for the plugin to reinitialise the configuration whether in debug mode or not.
+        System.setProperty("gmeFunctionalTest", "true");
```
dwalluck, line 119: Is this different from using `-D`? I guess without it it causes the need to restore system properties or at least I saw that added to some test classes in this PR.

rnc: Yes. If the tests are using debug they are running Gradle embedded and don't fork - hence the `-D` doesn't work and we need to use the current system properties. I provide both in case in future we fork a test and we want everything to still work (actually EHCache is still forked due to needing to use a different JDK).

PR #537, opened 2026-05-14
[NCL-9648] Apply shared workflow for build and snapshot

patrikk0123:


rnc:
Lets get #538 merged first as it should help simplify

rnc:
@patrikk0123 Apologies for the conflict due to #538 ; but it should help simplify https://github.com/project-ncl/shared-github-actions/pull/46

.github/workflows/gradle-pr.yml
```diff
@@ -0,0 +1,78 @@
...
+    paths-ignore: ["CONTRIBUTING.md", "LICENSE", "README.md"]
+
+jobs:
+  check-java-formatting:
+    uses: project-ncl/shared-github-actions/.github/workflows/gradle-ci.yml@<sha> # ver
+    with:
+      java_version: "17"
+      gradle_version: "8.3"
+      build_args: "--no-daemon --stacktrace spotlessJavaCheck"
+      remove_default_maven_settings: true
```
rnc: While I did wonder about moving the formatting check to gradle-ci like https://github.com/project-ncl/shared-github-actions/blob/main/.github/workflows/maven-ci.yml I guess the advantage of having it here is its run only once.

rnc:
This is looking good! I guess we need the shared-actions merged (and ideally released under a tag?) first?

patrikk0123:
@rnc tags applied

rnc (approved):

PR #551, opened 2026-06-04
[NCL-9648] Mend - fix exclusion path syntax

patrikk0123:


rnc:
@patrikk0123 Should this still have that SAST warning? What are the '3 neutral' checks?

patrikk0123:
@rnc The old Exclusion path was wrong. And current changes to workflow won't be applied to workflow run unless merged.

PR #558, opened 2026-06-08
Fix DifferentJVMTest to support macOS

cunningt:
## Summary
- `JVMTestSetup` was hardcoded for Linux (`x64_linux` archive filenames, no `Contents/Home` subdir), causing `DifferentJVMTest` to fail on macOS.
- Detect OS and architecture at class-load time so the correct Adoptium archive is downloaded on macOS — including `aarch64` for Apple Silicon on JDK 17 and `x64` via Rosetta for JDK 8 (where no aarch64 build exists).
- Resolve JDK home through `Contents/Home` on macOS to match the Adoptium tarball layout.

@rnc I don't have a Linux box - I wasn't able to test if anything regresses on Linux with this change - I'm hoping it runs cleanly on Linux with these changes.

## Test plan
- [x] Verified `DifferentJVMTest` passes on macOS
- [ ] Verify `DifferentJVMTest` passes on Linux (no regression — CI should cover this)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


rnc (approved):

rnc:
@cunningt I presume you have a Mac and the Mac side works for you?

cunningt:
@rnc Yes, Mac user here (and the tests work on OS X now)

PR #565, opened 2026-06-18
[NCL-9764] Fix CVEs

patrikk0123:


rnc (approved):

patrikk0123:
upgrading wiremock-jre8 to 2.35.2 won't do. will have to upgrade to v3

rnc:
Why is it complaining about a test dependency?

PR #574, opened 2026-07-02
Filter out unsupported JVM arguments from Gradle build environment

### LINKED ISSUE #571: Gradle environment / remove unsupported JVM arguments
A project's gradle.properties can include **org.gradle.jvmargs**, like https://github.com/spring-projects/spring-security/blob/5.7.11/gradle.properties#L9C1-L9C19

It'd be good to be able to strip these arguments when we're building with a JDK that no longer supports them.

cunningt:
When a project's gradle.properties includes org.gradle.jvmargs with options removed in later Java versions (e.g. -XX:MaxPermSize, -XX:+UseConcMarkSweepGC), the Gradle daemon fails with "Unrecognized VM option". Detect the target JDK version from its release file and filter out arguments that are no longer supported.

Fixes: https://github.com/project-ncl/gradle-manipulator/issues/571


rnc (approved):
Thanks!

PR #579, opened 2026-07-21
[NCL-9764] Fix CVEs

patrikk0123:


rnc:
This is an *excellent* and very neat change! It avoids the jackson problem entirely! Nice!

I've pushed a couple of changes to depend instead upon the versions from PME (and update PME to 5.5) to simplify the code.

rnc (approved):
Thanks!

PR #615, opened 2026-09-10
Update Gradle versions in gradle-pr.yml

rnc:


rnc:
Note : this is partially reverted in #619 due to problems with Gradle 8.14.5

PR #632, opened 2026-09-23
Replace Charset.defaultCharset() with StandardCharsets.UTF_8 in PluginUtils

koentsje:
The suggested fix for https://redhat.atlassian.net/browse/NCLSUP-1594


rnc:
@koentsje Thanks for the PR. Are you able to add tests please? I note also that we use `defaultCharset` a lot elsewhere as well so not sure whether those would also cause an issue. I have also put a question on the original JIRA.

